PR #9, opened 2013-10-30
add support for limiting job concurrency based on one or more build parameters

joemiller:
Submitting this PR to see if this feature set is something the upstream would be in favor of accepting. And if so, what changes or cleanup would be necessary to get it upstream?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.iml
 *.ipr
 *.iws
+.idea
 target
 work
 # eclipse =>
--- a/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
+++ b/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
@@ -30,7 +30,8 @@ import java.util.logging.Logger;
 public class ThrottleQueueTaskDispatcher extends QueueTaskDispatcher {
 
     @Override
-    public CauseOfBlockage canTake(Node node, Task task) {
+    public CauseOfBlockage canTake(Node node, Queue.BuildableItem item) {
+        Task task = item.task;
         if (task instanceof MatrixConfiguration) {
             return null;
         }
@@ -89,7 +90,7 @@ public class ThrottleQueueTaskDispatcher extends QueueTaskDispatcher {
         return null;
     }
 
-    // @Override on jenkins 4.127+ , but still compatible with 1.399
+    // @Override on jenkins 1.427+ , but still compatible with 1.399
     public CauseOfBlockage canRun(Queue.Item item) {
         ThrottleJobProperty tjp = getThrottleJobProperty(item.task);
         if (tjp!=null && tjp.getThrottleEnabled()) {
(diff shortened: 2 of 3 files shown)


varosi:
Do you plan to merge those changes about controlling throttleling based on build parameters?

jmozmoz:
I merge the current master branch into this pull request and tried to follow the review comments in this pull request. These I do not know how to implement:
- It also makes sense to use formatted outputs to prevent string constructions
- It makes sense to add the parameter validation to UI form. Otherwise, errors in data formats may be left unnoticed
- I suggest to calculate the array and save it to a transient field once (on save/load) to decrease the performance impact 
- Store a list instead of the string to avoid multiple parsings. I also think that the field name is quite confusing
- isAnotherBuildWithSameParametersRunningOnAnyNode() has its internal logging of conflicts (there seems to be no loggin in isAnotherBuildWithSameParametersRunningOnAnyNode)
- Please use Sun/Oracle Java code style to prevent complains from FindBugs and other utilities

If you give me more specific guidance how to fix these, I will try to implement them also.

Here is the repository: https://github.com/jmozmoz/throttle-concurrent-builds-plugin

jmozmoz:
Ping
Is there a chance to get update of the original pull request into this plugin? Anybody willing to help?

oleg-nenashev:
@varosi @jmozmoz 
This PR would be eligible for the merge when the performance issues get solved.
1.609.x LTS already provides some changes in this direction, but not enough IMHO.

Ignas:
Which performance issues exactly have to be solved? I can't find any of them being discussed in comments on this pull request.

jmozmoz:
As requested a I move the list paramsToCompare to a transient field which is filled by an overwritten load()/save()?
See the commit https://github.com/jmozmoz/throttle-concurrent-builds-plugin/commit/599453b44ee6423a2827be9d540f3ef48f019774 in https://github.com/pantheon-systems/throttle-concurrent-builds-plugin/pull/6

Is this the way to go?

jmozmoz:
As requested a I move the list paramsToCompare to a transient field which is filled by an overwritten load()/save()?
See the commit https://github.com/jmozmoz/throttle-concurrent-builds-plugin/commit/599453b44ee6423a2827be9d540f3ef48f019774 in https://github.com/pantheon-systems/throttle-concurrent-builds-plugin/pull/6

Is this the way to go?

oleg-nenashev:
@Ignas E.g. https://issues.jenkins-ci.org/browse/JENKINS-27650

oleg-nenashev:
@jmozmoz 
> As requested a I move the list paramsToCompare to a transient field which is filled by an overwritten load()/save()?

Yes, it may work. I've lost the track of this PR a bit, need to review it again. Did you have a chance to process the previous comments?

jmozmoz:
I created a new pull request #33 including upstream changes not merged into in this original pull request. Also I think I addressed the following issues (striked out):
- ~~I suggest to calculate the array and save it to a transient field once (on save/load) to decrease the performance impact~~ 
- ~~Store a list instead of the string to avoid multiple parsings. I also think that the field name is quite confusing~~ (this was already fixed (the list), also the string has been renamed)
- ~~isAnotherBuildWithSameParametersRunningOnAnyNode() has its internal logging of conflicts~~
- ~~Please use Sun/Oracle Java code style to prevent complains from FindBugs and other utilities~~
- It also makes sense to use formatted outputs to prevent string constructions
- It makes sense to add the parameter validation to UI form. Otherwise, errors in data formats may be left unnoticed
(the last two issues remain: Could somebody please point me to examples how to fix them)

jmozmoz:
Now the pull request #33 builds but there are some test failures. I have no idea, how these test failures are related to the work here. Could somebody please help.

zoidbergwill:
Can this PR be closed since https://github.com/jenkinsci/throttle-concurrent-builds-plugin/pull/38 has been merged?

PR #19, opened 2014-06-07
Comparison of Matrix configurations done by Matrix Root Project

papajulio:
This is a tentative fix to make work both options "Throttle this project alone" and "Throttle Matrix configuration builds" together.  

I don't know if this is completely correct, in my tests is working but I would like to get feedback before merging it and testing it thoroughly with all the possible combinations.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,7 @@ THE SOFTWARE.
   <artifactId>throttle-concurrents</artifactId>
   <packaging>hpi</packaging>
   <name>Jenkins Throttle Concurrent Builds Plug-in</name>
-  <version>1.8.2</version>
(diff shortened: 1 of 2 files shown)


oleg-nenashev:
@papajulio Yeh, the "next major version" has never happened :(

papajulio:
@oleg-nenashev yeah sorry, I was doing some cleaning. If you prefer to leave it open I don't have any problem but I thought this was dead. 

I did use this code for a year or so in my previous company and it worked ok :)

oleg-nenashev:
Yeah, something should change at some point.
I also changed the company and thus had no time to maintain all the stuff. Will reopen it just in case we find a new active maintainer. Maybe I will also find some time, but it is hard to commit

PR #20, opened 2014-06-09
JENKINS-12093 - Allow per-project & category throttling on the same job

gcummings:
Hi,

Please review and provide feedback.
Updated to always throttle on project settings, then optionally throttle by category.
Per existing behaviour, if project settings are 0, then no throttling occurs by project.

Other options could be
1. add new radio box to specify throttle by both project and category.
2. rework UI to have checkboxes to enable 'by project' and 'by category' independently so you can select both.

thanks
Geoff


raags:
This would be useful for us to apply both project and category level throttling. @oleg-nenashev @gcummings would you be able to resolve the conflicts and help merge this PR?

timotei:
@basil would there be availability to merge this PR if the conflicts are solved? We'd need something like this as well.

basil:
I'm not really maintaining this plugin anymore, you are free to adopt it as described in https://www.jenkins.io/doc/developer/plugin-governance/adopt-a-plugin/

PR #26, opened 2015-03-05
[WIP] Avoid looping over and over on Queue.pendings

ydubreuil:
Do not merge yet: work in progress

This patch tries to reduce contention on h.m.Queue lock by caching h.m.Queue.pendings content in a HashSet at the beginning of canTake / canRun, which is possible as we hold the lock at that time. HashSet  is then used to avoid call to Queue.isPending() which loops over Queue.pendings array for every check.

### DIFF
--- a/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
+++ b/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
@@ -14,6 +14,7 @@ import hudson.model.labels.LabelAtom;
 import hudson.model.queue.CauseOfBlockage;
 import hudson.model.queue.QueueTaskDispatcher;
 
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.logging.Level;
@@ -50,6 +51,9 @@ public class ThrottleQueueTaskDispatcher extends QueueTaskDispatcher {
                 }
             }
             else if (tjp.getThrottleOption().equals("category")) {
+                HashSet<Queue.BuildableItem> pendingItems = new HashSet<Queue.BuildableItem>(
+                        Hudson.getInstance().getQueue().getPendingItems());
+
                 // If the project is in one or more categories...
                 if (tjp.getCategories() != null && !tjp.getCategories().isEmpty()) {
                     for (String catNm : tjp.getCategories()) {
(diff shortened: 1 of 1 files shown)


jglick:
I hope to come up with a derivative PR soon. There is unfortunately a serious risk of conflict with an upcoming fix for [JENKINS-27708](https://issues.jenkins-ci.org/browse/JENKINS-27708).

oleg-nenashev:
Have you already started working on the JENKINS-27708 fix? Just to avoid
double efforts...
2 апр. 2015 г. 16:42 пользователь "Jesse Glick" <notifications@github.com>
написал:

jglick:
> Have you already started working on the JENKINS-27708 fix?

Me? No, I have no plans to touch that. I am hoping to offer a fix for JENKINS-27650.

jglick:
It seems that the change in this PR exacerbates [JENKINS-27708](https://issues.jenkins-ci.org/browse/JENKINS-27708) even in 1.580.2. I am not sure I see this point of this change anyway. It does replace a `List` search with `Set.contains`, which is generally a good idea, but the list is not generally very long to begin with, and the stated purpose—avoiding contention on the `Queue` lock—does not make sense to me, since `QueueTaskDispatcher` methods are called with the `Queue` lock held anyway (and modern JREs make it cheap to enter a `synchronized` method when the current thread already holds the monitor). By the same argument, though, I cannot figure out why it would introduce a race condition, since the result of `isPending` should not have been changing within a single call to `QueueTaskDispatcher`. So I am probably missing something. Unfortunately it is difficult to demonstrate whether the behavior at a given revision is right or wrong, because due to JENKINS-27650 once a few builds are started, the whole system becomes totally unresponsive and I cannot even see what is building.

jglick:
Specifically in my test environment, where the per-category maximum is set to 7, `canRun` computes a `totalRunCount` of 1 for a while—this seems to be the flyweight matrix parent, which IMO should not be counted at all, but whatever—and then suddenly starts computing 26 and thus blocking further builds. I would have expected this number to grow up to 7 and never higher. I suspect that the comment

```java
// I think this'll be more reliable than job.getBuilds(), which seemed to not always get
// a build right after it was launched, for some reason.
```

in `buildsOfProjectOnNode` means that this code is still not right, as suggested by JENKINS-27708. Perhaps the extreme environment of my test case just causes this bug to be manifested more reliably than it otherwise would. At any rate, that makes it hard to work on the performance, because I cannot tell if I am _introducing_ functional regressions.

oleg-nenashev:
I would say the plugin is totally broken for new core versions, so it may make sense to take a risk and to release any fix in a version with "-alpha" version and a new core dependency

ydubreuil:
> but the list is not generally very long to begin with

My assumption was the list could be quite long when multiple matrix jobs with a high number of configurations were pending. I wanted to reduce search time using a hash based lookup and ultimately reduce ThrottleQueueTaskDispatcher.canTake() execution time.

basil:
@ydubreuil The last update to this PR was over 4 years ago. Can this PR be closed to clean up the open pull requests list?

ydubreuil:
Not in progress, closing this out

PR #28, opened 2015-04-16
[WiP] - TCB plugin performance (extra caching)

oleg-nenashev:
The approach is not ready. 
* It provides some performance advantages on overloads in "categories" mode, but I'm still able to overload Jenkins. 
* Job configuration updates should be handled

Merge with #27 is also required

### DIFF
--- a/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
+++ b/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
@@ -24,6 +24,58 @@ import javax.annotation.Nonnull;
 @Extension
 public class ThrottleQueueTaskDispatcher extends QueueTaskDispatcher {
 
+    /**
+     * Fast and unreliable implementation of can take.
+     * It's designed to be a first-stage check in 
+     * {@link ThrottleQueueTaskDispatcher#canTake(hudson.model.Node, hudson.model.Queue.Task)}.
+     * @param node 
+     * @param task
+     * @param tjp Non-null job property of the task
+     * @return {@link CauseOfBlockage} if Jenkins cannot take the fob. 
+     *         null means that Jenkins <b>may</b> be able to take the job, but
+     *         a full check is required.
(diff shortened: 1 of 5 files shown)


src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
```diff
@@ -158,7 +233,7 @@ else if (tjp.getThrottleOption().equals("category")) {
                                 int totalRunCount = 0;
 
                                 for (AbstractProject<?,?> catProj : categoryProjects) {
-                                    if (Hudson.getInstance().getQueue().isPending(catProj)) {
+                                    if (checkPending && Hudson.getInstance().getQueue().isPending(catProj)) {
```
oleg-nenashev: It was a bad idea from another commit. I'll rework it

pom.xml
```diff
@@ -113,6 +113,12 @@ THE SOFTWARE.
             <version>2.0.1</version>
             <type>jar</type>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>matrix-project</artifactId>
+            <version>1.0</version>
+            <type>jar</type>
```
jglick: `type` can be omitted BTW

jglick:
Merging with #27 will be hard. Probably you could just merge the first three commits, but the fourth (`Snapshot`) would need to be recreated.

This seems risky: code mistakes or oversights could easily lead to a stale cache.

oleg-nenashev:
> This seems risky: code mistakes or oversights could easily lead to a stale cache.

Yes, I'm also aware about it. I consider flushing the cache on updates and other conditions. There's still a risk of false-negatives deadlock if something is implemented incorrectly. Probably, a periodic cache cleanup (e.g. every 1 minute) could prevent it

basil:
@oleg-nenashev The last update to this PR was 4 years ago. Can this PR be closed to clean up the open pull requests list?

PR #30, opened 2015-06-25
Acceptance Test for categories

wolfs:
I added two test to verify the behaviour of the throttle-concurrent-builds plugin. If these are accepted I will try to add some caching.
If you try these tests against 1.607 they will fail and expose JENKINS-27708.
If you want to have a different core version, then I am fine with that, too. Please tell me which one to use.
If you think there should be additional tests then tell me and I will try to add them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,7 +26,7 @@ THE SOFTWARE.
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.424</version>
+    <version>1.580.1</version>
   </parent>
 
   <artifactId>throttle-concurrents</artifactId>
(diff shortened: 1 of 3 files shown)


oleg-nenashev:
@wolfs 
Thanks for implementing the tests.
Caching for versions till 1.609.1 is quite complex6 because there's no reliable extension point to hook on. In 1.609 it's possible to somehow use Snapshot updates to update caches, but actually it would be better to enhance Jenkins internals somehow.

wolfs:
Can this be merged? Is still something missing?

oleg-nenashev:
LGTM :+1:

pom.xml
```diff
@@ -26,7 +26,7 @@ THE SOFTWARE.
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.424</version>
+    <version>1.580.1</version>
```
oleg-nenashev: Which is the minimal required version? Currently the plugin works well on 1.424.x..1.532.x and 1.609.x LTS lines only, so bumping the version may impact the existing installations. In particular I'd prefer to support 1.509.3 at least

wolfs: I was able to make it work with 1.554. For earlier version running with the newer guava version required for JGiven makes the other tests fail.

PR #31, opened 2015-08-15
Acceptance Test for categories

wolfs:
I added two test to verify the behaviour of the throttle-concurrent-builds plugin. If these are accepted I will try to add some caching.
If you try these tests against 1.607 they will fail and expose JENKINS-27708.
If you want to have a different core version, then I am fine with that, too. Please tell me which one to use.
If you think there should be additional tests then tell me and I will try to add them.

--
This is pull request #30 again. I accidently pushed to the master. I reverted my changes via a force-push 5 minutes later...

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -26,7 +26,7 @@ THE SOFTWARE.
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.580.1</version>
+    <version>1.554</version>
   </parent>
 
   <artifactId>throttle-concurrents</artifactId>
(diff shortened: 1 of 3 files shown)


oleg-nenashev:
@wolfs 
I'll contact my past colleagues regarding the version upgrade. If they are OK with the dependency upgrade above 1.509.x, I'll be ok with 1.532 (no executors calculation issue) and probably with the current version.

Do you consider also creating such pull request to https://github.com/jenkinsci/acceptance-test-harness?

oleg-nenashev:
@abayer 
So I propose to bump the plugin to 1.625.x directly.
1.532.x is pretty old (last one having stable TCB), and 1.625.x is the first version with stable queue after 1.609.1

oleg-nenashev:
Self-assigning, because it can be integrated now 1.609.3 is the current baseline

PR #32, opened 2015-11-11
Reduce applicability requirements

dsavenko:
Hi,

We have a project type, which does not extend AbstractProject, but directly extends Job and implements Queue.Task (we have certain reasons to not extend AbstractProject). We would like to have the Throttle plugin working with it, but it's tied to AbstractProject. 

This pull request unties it from AbstractProject class and makes so, that the Throttle plugin works with any class which extends Job and implements Queue.Task. 

This is just a technical patch, no new features added.

### DIFF
--- a/src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
+++ b/src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
@@ -1,10 +1,7 @@
 package hudson.plugins.throttleconcurrents;
 
 import hudson.Extension;
-import hudson.Util;
-import hudson.matrix.MatrixBuild;
 import hudson.matrix.MatrixConfiguration;
-import hudson.matrix.MatrixProject;
 import hudson.model.AbstractDescribableImpl;
 import hudson.model.Descriptor;
 import hudson.model.Item;
@@ -15,13 +12,10 @@ import hudson.model.JobPropertyDescriptor;
 import hudson.model.Queue;
 import hudson.util.FormValidation;
 import hudson.util.ListBoxModel;
-import jenkins.model.Jenkins;
-import net.sf.json.JSONObject;
-import org.kohsuke.stapler.DataBoundConstructor;
-import org.kohsuke.stapler.QueryParameter;
-import org.kohsuke.stapler.StaplerRequest;
+import hudson.Util;
+import hudson.matrix.MatrixBuild;
+import hudson.matrix.MatrixProject;
 
-import javax.annotation.CheckForNull;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -30,11 +24,19 @@ import java.util.List;
 import java.util.Map;
 import java.util.WeakHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
+import javax.annotation.CheckForNull;
+import jenkins.model.Jenkins;
+
+import net.sf.json.JSONObject;
+
+import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.QueryParameter;
+import org.kohsuke.stapler.StaplerRequest;
 
 public class ThrottleJobProperty extends JobProperty<Job<?,?>> {
     // Moving category to categories, to support, well, multiple categories per job.
     @Deprecated transient String category;
-
+    
     private Integer maxConcurrentPerNode;
     private Integer maxConcurrentTotal;
     private List<String> categories;
@@ -48,7 +50,7 @@ public class ThrottleJobProperty extends JobProperty<Job<?,?>> {
      * functionality upgrades.
      */
     private Long configVersion;
-
+    
     @DataBoundConstructor
     public ThrottleJobProperty(Integer maxConcurrentPerNode,
                                Integer maxConcurrentTotal,
@@ -94,19 +96,19 @@ public class ThrottleJobProperty extends JobProperty<Job<?,?>> {
             }
         }
         configVersion = 1L;
-
+        
         // Handle the throttleConfiguration in custom builds (not released)
         if (throttleConfiguration && matrixOptions == null) {
             matrixOptions = new ThrottleMatrixProjectOptions(false, true);
         }
-
+        
         return this;
     }
 
     @Override protected void setOwner(Job<?,?> owner) {
         super.setOwner(owner);
         if (throttleEnabled && categories != null) {
-            DescriptorImpl descriptor = (DescriptorImpl) getDescriptor();
+            DescriptorImpl descriptor = (DescriptorImpl) getDescriptor();    
             synchronized (descriptor.propertiesByCategoryLock) {
                 for (String c : categories) {
                     Map<ThrottleJobProperty,Void> properties = descriptor.propertiesByCategory.get(c);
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
```diff
@@ -24,20 +30,11 @@
...
-import org.kohsuke.stapler.DataBoundConstructor;
-import org.kohsuke.stapler.QueryParameter;
-import org.kohsuke.stapler.StaplerRequest;
-
-public class ThrottleJobProperty extends JobProperty<AbstractProject<?,?>> {
+public class ThrottleJobProperty extends JobProperty<Job<?,?>> {
     // Moving category to categories, to support, well, multiple categories per job.
     @Deprecated transient String category;
-    
+
```
oleg-nenashev: Please avoid such formatting changes to simplify merges

src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
```diff
@@ -24,20 +30,11 @@
...
-import jenkins.model.Jenkins;
 
-import net.sf.json.JSONObject;
-
-import org.kohsuke.stapler.DataBoundConstructor;
-import org.kohsuke.stapler.QueryParameter;
-import org.kohsuke.stapler.StaplerRequest;
-
-public class ThrottleJobProperty extends JobProperty<AbstractProject<?,?>> {
+public class ThrottleJobProperty extends JobProperty<Job<?,?>> {
```
oleg-nenashev: May be a binary compatibility issue, I need to check it

src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
```diff
@@ -180,10 +180,10 @@ else if (tjp.getThrottleOption().equals("category")) {
 
     @CheckForNull
     private ThrottleJobProperty getThrottleJobProperty(Task task) {
-        if (task instanceof AbstractProject) {
-            AbstractProject<?,?> p = (AbstractProject<?,?>) task;
+        if (task instanceof Job) {
+            Job<?,?> p = (Job<?,?>) task;
             if (task instanceof MatrixConfiguration) {
-                p = (AbstractProject<?,?>)((MatrixConfiguration)task).getParent();
+                p = (Job<?,?>)((MatrixConfiguration)task).getParent();
```
oleg-nenashev: BTW class cast to Job should not be required here

src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
```diff
@@ -228,7 +228,7 @@ private int buildsOfProjectOnAllNodes(Task task) {
     private int buildsOnExecutor(Task task, Executor exec) {
         int runCount = 0;
         if (exec.getCurrentExecutable() != null
-            && exec.getCurrentExecutable().getParent() == task) {
+            && task.equals(exec.getCurrentExecutable().getParent())) {
```
oleg-nenashev: This replacement is suspicious. Why did you need it?

src/test/java/hudson/plugins/throttleconcurrents/ThrottleJobPropertyTest.java
```diff
@@ -1,8 +1,7 @@
 package hudson.plugins.throttleconcurrents;
 
-import hudson.model.AbstractProject;
-import hudson.model.FreeStyleProject;
-import hudson.model.Job;
+import hudson.model.*;
```
oleg-nenashev: It's a bad practice in the case of the non-test code

src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
dsavenko: Sure, I will get rid of them. Actually, I removed them manually before commit, but IDE put them again :)

src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
dsavenko: A couple of reasons:
1. Our job class has equals/hashCode overriden
2. In Jenkins core, they use equals for similar purposes, so I assume, that it is OK. E.g. here's a chunk of code from the hudson.model.Queue class:

    public boolean cancel(Task p) {
        lock.lock();
        try { try {
            LOGGER.log(Level.FINE, "Cancelling {0}", p);
            for (WaitingItem item : waitingList) {
                if (item.task.equals(p)) {
                    return item.cancel(this);
                }
            }
            // use bitwise-OR to make sure that both branches get evaluated all the time
            return blockedProjects.cancel(p) != null | buildables.cancel(p) != null;
        } finally { updateSnapshot(); } } finally {
            lock.unlock();
        }
    }

3. It's generally considered a good practice to use equals() instead of reference equality checks.

src/test/java/hudson/plugins/throttleconcurrents/ThrottleJobPropertyTest.java
dsavenko: Sure, I will change it.

jglick:
You probably want to implement `ParameterizedJobMixIn.ParameterizedJob` and have TCB look for that instead.

dsavenko:
Good point, I haven't thought about it. Will provide the update soon.

dsavenko:
@jglick, hmm... On second thought, ParameterizedJobMixIn is there since 1.556, while TCB is built against 1.424. Also, JobProperty's generic parameter must extend Job<?, ?>, so I'm bound to extend the ThrottleJobProperty class from JobProperty<Job<?,?>> (can't use ParameterizedJobMixIn.ParameterizedJob, as it doesn't extend Job).

oleg-nenashev:
Personally I don't like the <code>SometihgMixIn</code> approach much, because it mostly workarounds Java's inheritance mechanism. It does not mean I can propose a better solution...

<code>ParameterizedJobMixIn.ParameterizedJob</code> is also not guaranteed to be used by all Job implementations, so I'd prefer the current approach.

In any case, the core dependency update may be reasonable. Versions 1.535...1.609.2 do not behave reliably, so probably bumping to 1.609.2 is the best option. The plugin needs some redesign, so please do not consider the version dependency as a blocker.

dsavenko:
So, what should I do next?

abayer:
Dependency update is perfectly fine, yeah.

src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
oleg-nenashev: Agreed

abayer:
fwiw, tested this and it works fine throttling groups of Workflow jobs - not being used within the Workflow script, though. That's more work, which I'll get diving into once this is in.

oleg-nenashev:
:+1: for the current implementation

abayer:
Rightie-o, then I'll merge it.

dsavenko:
Thanks for merging!

PR #34, opened 2016-02-11
Throttling flyweight tasks

dsavenko:
Hi,

This pull request enables throttling flyweights tasks. No new UI's added, but now if a user turns on throttling for a flyweight job, it works. 

Some bits of explanation:

1. Throttling logic is not changed. It's only moved to the Throttler class. This is done because I need it in two places.
2. Since QueueTaskDispatcher is not called for flyweights, I use the NodeProperty's canTake() instead. There is a gap between returning from canTake() and the build being assigned to some node (leave the queue). Another throttle subject may come to this node in this gap, which can result in throttling condition being violated. This is why I needed a static lock. With this lock, I make sure we don't make any actual decisions about throttling until the previous build actually leaves the queue. 
3. I need ThrottleNodeProperty to be present in all nodes, so manually add it to each node on every computer change. 
4. Behaviour is not changed for non-flyweight tasks.
5. I changed the Jenkins target version to 1.609.3, since earlier we agreed it's OK (see https://github.com/jenkinsci/throttle-concurrent-builds-plugin/pull/32).

I realise the implementation's somewhat unusual. But I spent some time studying the Jenkins code and didn't found any other options to throttle flyweight tasks, and we really need to do it. 

Please, share your thoughts.


oleg-nenashev:
needs a review

sshelomentsev:
Hi @damphyr 
Yes, this PR solve the problem with throttling concurrent jobs in the pipeline.

oleg-nenashev:
I doubt it completely solves the throttling topic, at least for <code>parallel()</code> pipeline invocations. In general, this PR definitely deserves some attention from maintainers and integration. Node throttling approach rework is also reasonable.

Sorry about dropping a ball on it.

yarps95:
Hi, I've merged master into our branch and made some minor improvements. Could you please review it?

mamh2021:
any update???

ggkochanski:
@damphyr can you check if adding quietPeriod to build() doesn't help accidently?
e.g.
`build job: 'Compile', quietPeriod: 3,  parameters: [[$class: 'StringParameterValue', name: 'TARGET', value: region]]`

tj-cisco:
Any updates on this merge? Seems like it would fix the race condition where throttling is violated when using upstream project triggers:

```
        pipelineTriggers {
            triggers {
                upstream {
                    upstreamProjects('triggeringProject')
                    threshold('SUCCESS')
                }
            }
        }
```

PR #35, opened 2016-02-14
[JENKINS-25326] Make work with cloudbees-folder with permissions.

ikedam:
[JENKINS-25326](https://issues.jenkins-ci.org/browse/JENKINS-25326)

Throttle-concurrent-builds doesn't work with projects inside cloudbees-folder and not accessible from anonymous, as the queue thread work as anonymous.

This change elevates throttle-concurrent-builds to SYSTEM when accessing projects.

### DIFF
--- a/src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
+++ b/src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
@@ -10,6 +10,8 @@ import hudson.model.Job;
 import hudson.model.JobProperty;
 import hudson.model.JobPropertyDescriptor;
 import hudson.model.Queue;
+import hudson.security.ACL;
+import hudson.security.NotSerilizableSecurityContext;
 import hudson.util.FormValidation;
 import hudson.util.ListBoxModel;
 import hudson.Util;
@@ -29,6 +31,8 @@ import jenkins.model.Jenkins;
 
 import net.sf.json.JSONObject;
 
+import org.acegisecurity.context.SecurityContext;
+import org.acegisecurity.context.SecurityContextHolder;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.QueryParameter;
 import org.kohsuke.stapler.StaplerRequest;
@@ -199,10 +203,14 @@ public class ThrottleJobProperty extends JobProperty<Job<?,?>> {
         return categoryTasks;
     }
     private static Item getItem(ItemGroup group, String name) {
-        if (group instanceof Jenkins) {
-            return ((Jenkins) group).getItemMap().get(name);
-        } else {
+        SecurityContext orig = SecurityContextHolder.getContext();
+        NotSerilizableSecurityContext auth = new NotSerilizableSecurityContext();
+        auth.setAuthentication(ACL.SYSTEM);
+        SecurityContextHolder.setContext(auth);
+        try {
             return group.getItem(name);
+        } finally {
+            SecurityContextHolder.setContext(orig);
         }
     }
     
--- a/src/test/java/hudson/plugins/throttleconcurrents/ThrottleJobPropertyTest.java
+++ b/src/test/java/hudson/plugins/throttleconcurrents/ThrottleJobPropertyTest.java
@@ -12,7 +12,9 @@ import org.acegisecurity.context.SecurityContextHolder;
 import org.jvnet.hudson.test.Bug;
 import org.jvnet.hudson.test.HudsonTestCase;
 
+/*
 import com.cloudbees.hudson.plugins.folder.Folder;
+*/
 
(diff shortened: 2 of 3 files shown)


src/test/java/hudson/plugins/throttleconcurrents/ThrottleJobPropertyTest.java
```diff
@@ -124,27 +139,21 @@ public void testDescriptorImpl_should_a_concurrency_safe_list_for_categories(){
...
-            return new AuthorizationStrategy.Unsecured().getRootACL();
-        }
-        @Override public Collection<String> getGroups() {
-            return Collections.emptySet();
-        }
-        @Override public ACL getACL(Job<?,?> project) {
-            fail("not even supposed to be looking at " + project);
-            return super.getACL(project);
-        }
-    }
```
ikedam: Existing tests used custom `AuthorizationStrategy` to ensure `getACL` is not used.
This is not applicable to cloudbees-folder as it checks `READ` permissions.
Anyway, that's not actually what we want to test, but what we want to test is that throttle-consurrent-build works even as `ANONYMOUS` with no permissions.

ikedam:
The test fails as defcf134b55e0094e99b4fc6c02c959fa6ef23b1 only added a test to reproduce JENKINS-25326 and didn't fix anything.
https://jenkins.ci.cloudbees.com/job/plugins/job/throttle-concurrent-builds-plugin/org.jenkins-ci.plugins$throttle-concurrents/78/testReport/junit/hudson.plugins.throttleconcurrents/ThrottleJobPropertyTest/testGetCategoryProjectsInFolder/
```
junit.framework.AssertionFailedError: expected:<[hudson.model.FreeStyleProject@3048fb06[folder/p]]> but was:<[]>
	at junit.framework.Assert.fail(Assert.java:50)
	at junit.framework.Assert.failNotEquals(Assert.java:287)
	at junit.framework.Assert.assertEquals(Assert.java:67)
	at junit.framework.Assert.assertEquals(Assert.java:74)
	at hudson.plugins.throttleconcurrents.ThrottleJobPropertyTest.assertProjects(ThrottleJobPropertyTest.java:150)
	at hudson.plugins.throttleconcurrents.ThrottleJobPropertyTest.testGetCategoryProjectsInFolder(ThrottleJobPropertyTest.java:57)
```

It demonstrates throttle-concurent-build doesn't work with cloudbees-folder. `ThrottleJobProperty#getItem` doesn't return a project unreadable from anonymous and inside a cloudbees-folder.
I'll add a commit to fix it.

src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
```diff
@@ -199,10 +203,14 @@ public boolean isThrottleMatrixConfigurations() {
         return categoryTasks;
     }
     private static Item getItem(ItemGroup group, String name) {
-        if (group instanceof Jenkins) {
-            return ((Jenkins) group).getItemMap().get(name);
-        } else {
+        SecurityContext orig = SecurityContextHolder.getContext();
+        NotSerilizableSecurityContext auth = new NotSerilizableSecurityContext();
+        auth.setAuthentication(ACL.SYSTEM);
+        SecurityContextHolder.setContext(auth);
```
ikedam: `ACL#impersonate` is not available in Jenkins 1.424, which is targeted by throttle-concurrent-build.

|Method                                          |Introduced version|
|:-----------------------------------------------|:-----------------|
|`ACL#impersonate(Authentication)`               |1.462             |
|`ACL#impersonate(Authentication, Runnable)`     |1.509             |
|`ACL#impersonate(Authentication, Callable<V,T>)`|1.587             |

ikedam:
The test for couldbees-folder succeeds with 0299dac0e627ac5146102fada24f58fa31d9be99.

oleg-nenashev:
I doubt the queue works as anonymous. If yes, it smells like a bug. 

The current approach hammers performance due to multiple context switches.
I propose to change to impersonate once on the top level (ThrottleQueueTaskDispatcher).

ikedam:
> I doubt the queue works as anonymous. If yes, it smells like a bug.

As far as I tested, that looks since Jenkins 1.536.
I believe the changelog "Executor threads are now created only on demand." relates to this issue.

> The current approach hammers performance due to multiple context switches.
> I propose to change to impersonate once on the top level (ThrottleQueueTaskDispatcher).

I'm not sure setting security contexts actually causes performance issues.
I tried to minimize the code section executed with the SYSTEM authorization.

Anyway, OK, I'll try that.
I think the problem is that throttle-concurrent-builds-plugin doesn't have appropriate tests for the functionality of `ThrottleQueueTaskDispatcher`...

ikedam:
I'll try a different approach, no longer based on this request.
I abandon this request.

PR #36, opened 2016-02-20
Added tests to verify ThrottleJobProperty actually works for builds

ikedam:
No tests see `ThrottleJobProperty` actually works when running builds.
I introduced `ExecutorWaterMarkRetentionStrategy` which counts how many builds run on the node at the same time, and added tests to see `ThrottleJobProperty` controls the number of concurrent builds.

This is preparation for the work for [JENKINS-25326](https://issues.jenkins-ci.org/browse/JENKINS-25326), which requires tests to see how `ThrottleJobProperty` works when called from executor thread.


oleg-nenashev:
:+1:

oleg-nenashev:
Thanks a lot!

PR #37, opened 2016-02-21
[JENKINS-25326] Elevates to SYSTEM when handling throttling for builds

ikedam:
[JENKINS-25326](https://issues.jenkins-ci.org/browse/JENKINS-25326)

Throttling doesn't work for projects in cloudbees-folder with latest Jenkins.
As far as I tested, it doesn't work since Jenkins 1.536:

| Jenlkins | Throttling in folder | Note |
|:---------|:---------------------|:-----|
| 1.532    | Works                | |
| 1.535    | Works                | |
| 1.536    | **Doesn't work**     | |
| 1.554    | **Doesn't work**     | |
| 1.642    | **Doesn't work**     | Requires some changes in test codes to compile |

I fix that problem in followng steps in this request:

1. Add a commit to add a test to reproduce the problem. This requires to change the targeted Jenkins version to >= 1.536.
2. Add a commit to fix the problem.
3. Add a commit to revert changes to reproduce the problem, including the change of targeted Jenkins version.

PR #41, opened 2016-04-10
[JENKINS-25326] - Elevate user to system during build throttling

oleg-nenashev:
Integrates #37 

CC @ikedam


oleg-nenashev:
Tests passed, integrating it

ikedam:
Thanks!
Looks good to me.

PR #44, opened 2016-04-16
JENKINS-34294: Added support to throttle Inheritance projects.

J-cztery:
Added support to throttle builds of inheritance-project: 
- Example
  
  Project **D1** and **D2** inherit from project B which inherits from project **A** and: 
  1. project **A** has configured throttling for category _**C1**_  and project **B** has configured throttling for category **_C2_** 
  
  then: 
  
  Projects **D1** and **D2** will act as if they had configured category for **_C1_** and **_C2_**. 
- Example2:
  Project **D1** and **D2** inherit from project **B** which inherits from project **A** and: 
  1. project **A** has configured throttling for project and MaxConcurrentPerNode = 1,    MaxConcurrentTotal=4
  2. project **B** has configured throttling for project and MaxConcurrentPerNode = 2, MaxConcurrentTotal=3
  
  then: 
  
  Projects **D1** and **D2** will act as if they had configured **_MaxConcurrentPerNode = 1, MaxConcurrentTotal=3_**. Projects **D1** and **D2** will be independent.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,7 +107,12 @@ THE SOFTWARE.
             <artifactId>matrix-project</artifactId>
             <version>1.4.1</version>
         </dependency>
-        
+        <dependency>
+            <groupId>hudson.plugins</groupId>
+            <artifactId>project-inheritance</artifactId>
+            <version>1.5.3</version>
+            <optional>true</optional>
+        </dependency>
         <!-- Dependencies for test -->
         <dependency>
           <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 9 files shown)


J-cztery:
Work in progress. Unit tests are planned, better handling of project throttling. Testing and fixing   parameter based throttling. But general solution idea should be clear.

oleg-nenashev:
@J-cztery Sorry for not responding for a long time. Is it ready for review now? Or still WiP?

J-cztery:
@oleg-nenashev : It is ok. I would appreciate a quick review. Unit tests are passing when i just go mvn test but for some reason they do not pass when i go install. If you agree that general direction is good, i will finish it off.

oleg-nenashev:
Oops. I missed the notification.
I also moved on in 2016 or so :(

PR #46, opened 2017-03-08
[JENKINS-31801] Add Pipeline throttle(category) step

abayer:
[JENKINS-31801](https://issues.jenkins-ci.org/browse/JENKINS-31801)

- [x] POC
- [x] Getting to finalized implementation
- [x] Snippet generator support
- [x] Tests!

Note that yes, this depends on `workflow-api` - I need that for `FlowExecution` and friends. And I had to remove the existing `guava:18.0` test dependency due to that violently conflicting with `StepContext`. I haven't actually run tests yet, let alone written new ones, but I wanted to throw this up for initial feedback ASAP.

cc @reviewbybees esp @jglick for Pipeline sanity review and @oleg-nenashev for Throttle sanity review


pom.xml
```diff
@@ -44,10 +44,10 @@ THE SOFTWARE.
   </licenses>
     
   <properties>
-    <jenkins.version>1.609.3</jenkins.version>
+    <jenkins.version>1.642.3</jenkins.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
jglick: delete

pom.xml
```diff
@@ -44,10 +44,10 @@ THE SOFTWARE.
...
-    <compileSource>1.6</compileSource>
-    <compileTarget>1.6</compileTarget>
+    <compileSource>1.7</compileSource>
+    <compileTarget>1.7</compileTarget>
```
jglick: delete

pom.xml
```diff
@@ -44,10 +44,10 @@ THE SOFTWARE.
   </licenses>
     
   <properties>
-    <jenkins.version>1.609.3</jenkins.version>
+    <jenkins.version>1.642.3</jenkins.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-    <compileSource>1.6</compileSource>
-    <compileTarget>1.6</compileTarget>
+    <compileSource>1.7</compileSource>
```
jglick: delete

src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
```diff
@@ -229,29 +243,82 @@ public boolean isThrottleMatrixConfigurations() {
         assert category != null && !category.equals("");
         List<Queue.Task> categoryTasks = new ArrayList<Queue.Task>();
         Collection<ThrottleJobProperty> properties;
-        DescriptorImpl descriptor = Jenkins.getActiveInstance().getDescriptorByType(DescriptorImpl.class);    
+        DescriptorImpl descriptor = Jenkins.getActiveInstance().getDescriptorByType(DescriptorImpl.class);
```
jglick: Please avoid gratuitous whitespace changes. Configure your editor to not make them.

src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
```diff
@@ -229,29 +243,82 @@ public boolean isThrottleMatrixConfigurations() {
...
+
+        DescriptorImpl descriptor = Jenkins.getActiveInstance().getDescriptorByType(DescriptorImpl.class);
+        for (Map.Entry<String,Integer> currentPipeline : descriptor.getThrottledPipelinesForCategory(category).entrySet()) {
+            Run<?,?> flowNodeRun = Run.fromExternalizableId(currentPipeline.getKey());
+
+            if (flowNodeRun == null) {
+                // No run found, so remove the throttle.
+                descriptor.removeThrottledPipelineForCategory(currentPipeline.getKey(), category, null);
+            } else if (!(flowNodeRun instanceof FlowExecutionOwner.Executable)) {
+                // If for some reason we've somehow ended up with a non-pipeline job, remove the throttle.
```
jglick: Why not just index by `FlowExecutionOwner` to begin with?

src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
```diff
@@ -229,29 +243,82 @@ public boolean isThrottleMatrixConfigurations() {
...
+                                        }
+                                        throttledPipelines.add(info);
+                                    }
+                                }
+                            } catch (IOException e) {
+                                // TODO: What do we do here if anything?
+                            }
+                        }
+                    } catch (InterruptedException | ExecutionException e) {
+                        // TODO: What do we do here if anything?
```
jglick: If you are calling `Future.get` that is a bad sign. Means you are causing this thread, whatever it is, to block on the CPS VM thread, and potentially to even wait indefinitely for a build to rehydrate its pickles.

Use a callback on the future instead, which means refactoring everything else to be asynchronous.

src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
```diff
@@ -229,29 +243,82 @@ public boolean isThrottleMatrixConfigurations() {
...
+                descriptor.removeThrottledPipelineForCategory(currentPipeline.getKey(), category, null);
+            } else {
+                FlowExecutionOwner owner = ((FlowExecutionOwner.Executable)flowNodeRun).asFlowExecutionOwner();
+                FlowExecution execution = owner.getOrNull();
+                if (execution == null) {
+                    // For some reason, the flow execution is null, so again? Remove the throttle.
+                    descriptor.removeThrottledPipelineForCategory(currentPipeline.getKey(), category, null);
+                } else {
+                    try {
+                        for (StepExecution stepExec : execution.getCurrentExecutions(false).get()) {
```
jglick: I am not sure why you think you need to do this but it has a very bad smell.

If you really have no `FlowNode` or `StepContext` that you are starting with, probably you want to pass `true`, since the innermost step should have the `Computer` in context if it somewhere inside `node`.

But the fact that you do _not_ have a specific step in mind is alarming. What is this code doing when you have a dozen `node` blocks running in `parallel`?

Did you rather mean to pay attention to the actual `ExecutorStepExecution.PlaceholderTask` being scheduled and look up its `context`? Currently there is no supported API for extracting contextual objects out of such a `Task`, so we would need to define one (cf. discussion in https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/26).

src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
```diff
@@ -262,7 +329,11 @@ private static Item getItem(ItemGroup group, String name) {
     
     @Extension
     public static final class DescriptorImpl extends JobPropertyDescriptor {
+        private static final Logger LOGGER = Logger.getLogger(DescriptorImpl.class.getName());
+
         private List<ThrottleCategory> categories;
+
+        private Map<String,Map<String,Integer>> throttledPipelines;
```
jglick: as above, you could just use `FlowExecutionOwner` as the key

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStep.java
```diff
@@ -0,0 +1,28 @@
...
+    public String getCategory() {
+        return category;
+    }
+
+    @Override
+    public StepExecution start(StepContext context) throws Exception {
+        return new ThrottleStepExecution(this, context);
+    }
+
+}
```
jglick: No descriptor?

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStepExecution.java
```diff
@@ -0,0 +1,69 @@
...
+
+    @Override
+    public boolean start() throws Exception {
+        ThrottledStepInfo info = new ThrottledStepInfo(getCategory());
+
+        ThrottledStepInfo parentInfo = getContext().get(ThrottledStepInfo.class);
+        if (parentInfo != null) {
+            // Make sure we record the node used for the parent throttle step if it exists.
+            Computer computer = getContext().get(Computer.class);
+            if (computer != null && computer.getNode() != null) {
```
jglick: Return `Node` from `getRequiredContext` and skip both the null check and the `getNode` conversion.

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStepExecution.java
```diff
@@ -0,0 +1,69 @@
...
+        body = getContext().newBodyInvoker()
+                .withContext(info)
+                .withCallback(BodyExecutionCallback.wrap(getContext()))
+                .start();
+        return false;
+    }
+
+    @Override
+    public void stop(Throwable cause) throws Exception {
+        if (body != null)
```
jglick: This method can be left blank—will not be called for a block-scoped step anyway, unless it is doing something between returning from `start` and the `BodyInvoker.start`.

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStepExecution.java
```diff
@@ -0,0 +1,69 @@
...
+        Run<?, ?> r = getContext().get(Run.class);
+        TaskListener listener = getContext().get(TaskListener.class);
+
+        ThrottleJobProperty.DescriptorImpl descriptor = Jenkins.getActiveInstance().getDescriptorByType(ThrottleJobProperty.DescriptorImpl.class);
+
+        descriptor.addThrottledPipelineForCategory(r.getExternalizableId(), getCategory(), listener);
+
+        body = getContext().newBodyInvoker()
+                .withContext(info)
+                .withCallback(BodyExecutionCallback.wrap(getContext()))
```
jglick: I think you meant to use `TailCall` and call `removeThrottledPipelineForCategory`.

src/main/resources/hudson/plugins/throttleconcurrents/Messages.properties
```diff
@@ -3,4 +3,6 @@ ThrottleQueueTaskDispatcher.MaxCapacityTotal=Already running {0} builds across a
 ThrottleQueueTaskDispatcher.BuildPending=A build is pending launch
 ThrottleQueueTaskDispatcher.OnlyOneWithMatchingParameters=A build with matching parameters is already running
 
-ThrottleMatrixProjectOptions.DisplayName=Additional options for Matrix projects
\ No newline at end of file
+ThrottleMatrixProjectOptions.DisplayName=Additional options for Matrix projects
+
+ThrottleJobProperty.DescriptorImpl.NoSuchCategory=Requested category "{0}" does not exist, so cannot throttle.
```
jglick: trailing newline

pom.xml
abayer: Yup yup - at one point, I was experimenting with stuff that did need newer versions of other dependencies and bumped things, but I'll roll this back now.

src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
abayer: Is there an easy way to go from some easily serializable identifier for `FlowExecutionOwner` to an actual instance of `FlowExecutionOwner`, like with `Run.fromExternalizableId`?

src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
abayer: Ok, I'll dig into that.

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStep.java
abayer: whoops.

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStepExecution.java
abayer: Aaaaah, for some reason I didn't think that was possible. Win!

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStepExecution.java
abayer: Aaaah, you're right. Thanks.

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStepExecution.java
abayer: Okiedokie.

abayer:
@jglick Did a fairly drastic rewrite here - I think I got out of the blocking-CPS-VM issues, and am tracking `FlowNode`s of the start nodes for the step, but I'm not sure if I'm actually on the right path. =)

casz:
I would be happy to test this! 👍

casz:
Would this be a correct usage, see test stage?
I might rewrite it to scripted, to get parallel with failfast!

Sorry for sick indentation but such is life 🍎 

```groovy
pipeline {
    agent none
    options {
        buildDiscarder(logRotator(numToKeepStr:'20'))
    }
    stages {
        stage('Build') {
            agent{ label 'ios '}
            steps {
                withCredentials([string(credentialsId: 'keychain', variable: 'PASSWORD')]) {
                    sh "security list-keychains -s ~/Library/Keychains/login.keychain"
                    sh "security unlock-keychain -p ${PASSWORD} ~/Library/Keychains/login.keychain"
                }
                wrap([$class: 'AnsiColorBuildWrapper', 'colorMapName': 'XTerm', 'defaultFg': 1, 'defaultBg': 2]) {
                    withCredentials([
                        string(credentialsId: 'fastlaneMatch', variable: 'MATCH_PASSWORD'),
                        string(credentialsId: 'fastlanePassword', variable: 'FASTLANE_PASSWORD')])
                    {
                        sh 'fastlane test'
                    }
                }
            }
        }
        stage('Test') {
            agent none
            steps {
                throttle('ios-simulator') {
                    node('ios') {
                        lock("ios-simulator-${env.NODE_NAME}") { // Do I even need lock?
                            withCredentials([string(credentialsId: 'keychain', variable: 'PASSWORD')]) {
                                sh "security list-keychains -s ~/Library/Keychains/login.keychain"
                                sh "security unlock-keychain -p ${PASSWORD} ~/Library/Keychains/login.keychain"
                            }
                            retry(5) {
                                wrap([$class: 'AnsiColorBuildWrapper', 'colorMapName': 'XTerm', 'defaultFg': 1, 'defaultBg': 2]) {
                                    withCredentials([
                                        string(credentialsId: 'fastlaneMatch', variable: 'MATCH_PASSWORD'),
                                        string(credentialsId: 'fastlanePassword', variable: 'FASTLANE_PASSWORD')])
                                    {
                                        sh 'fastlane test'
                                    }
                                }
                            }
                        }
                    }
                }
            }
        }
    }
}
```

pom.xml
```diff
@@ -26,7 +26,7 @@ THE SOFTWARE.
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.6</version>
+    <version>2.22</version>
```
abayer: 2.23 led to errors due to a `TestExtension` usage in `workflow-support`'s tests, so I bumped back down to 2.22 for now at least.

pom.xml
```diff
@@ -44,10 +44,10 @@ THE SOFTWARE.
   </licenses>
     
   <properties>
-    <jenkins.version>1.609.3</jenkins.version>
+    <jenkins.version>1.642.3</jenkins.version>
```
abayer: As mentioned in the commit message, we needed to bump dependency versions to get the latest `durable-task-step` for `PlaceholderTask.getNode()`.

pom.xml
abayer: Brought these back now that we're depending on 1.642.3 again.

src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
```diff
@@ -61,83 +70,102 @@ public CauseOfBlockage canTake(Node node, Task task) {
     private CauseOfBlockage canTakeImpl(Node node, Task task) {
         final Jenkins jenkins = Jenkins.getActiveInstance();
         ThrottleJobProperty tjp = getThrottleJobProperty(task);
-        
+        List<String> pipelineCategories = categoriesForPipeline(task);
+
         // Handle multi-configuration filters
-        if (!shouldBeThrottled(task, tjp)) {
+        if (!shouldBeThrottled(task, tjp) && pipelineCategories.isEmpty()) {
```
abayer: So re: this and later similar ugliness - I plan to rewrite `ThrottleQueueTaskDispatcher` pretty much completely before I'm done here, or at least rationalize it a lot better. This was me wedging things together until they worked, with clean up planned for afterwards.

src/test/java/hudson/plugins/throttleconcurrents/ThrottleStepTest.java
```diff
@@ -0,0 +1,122 @@
...
+                setupAgentsAndCategories();
+                WorkflowJob firstJob = story.j.jenkins.createProject(WorkflowJob.class, "first-job");
+                // This should be sandbox:true, but when I do that, I get org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use method groovy.lang.GroovyObject invokeMethod java.lang.String java.lang.Object
+                // And I cannot figure out why. So for now...
+                firstJob.setDefinition(new CpsFlowDefinition("throttle('" + ONE_PER_NODE + "') {\n" +
+                        "  echo 'hi there'\n" +
+                        "  node('first-agent') {\n" +
+                        "    semaphore 'wait-first-job'\n" +
+                        "  }\n" +
+                        "}\n", false));
```
abayer: Soooo - as I mentioned a few lines up, I'm getting script security exceptions on this when the sandbox is enabled and I can't figure out why, so for the moment, disabling script security until someone (probably @jglick!) tells me what I'm doing wrong to cause those errors to show up. =)

src/test/java/hudson/plugins/throttleconcurrents/ThrottleStepTest.java
```diff
@@ -0,0 +1,278 @@
...
+
+    private CpsFlowDefinition getJobFlow(String jobName, String category, String label) {
+        // This should be sandbox:true, but when I do that, I get org.jenkinsci.plugins.scriptsecurity.sandbox.RejectedAccessException: Scripts not permitted to use method groovy.lang.GroovyObject invokeMethod java.lang.String java.lang.Object
+        // And I cannot figure out why. So for now...
+        return new CpsFlowDefinition("throttle('" + category + "') {\n" +
+                "  echo 'hi there'\n" +
+                "  node('" + label + "') {\n" +
+                "    semaphore 'wait-" + jobName + "-job'\n" +
+                "  }\n" +
+                "}\n", false);
```
abayer: Soooo - as I mentioned a few lines up, I'm getting script security exceptions on this when the sandbox is enabled and I can't figure out why, so for the moment, disabling script security until someone (probably @jglick!) tells me what I'm doing wrong to cause those errors to show up. =)

casz:
Does this supersede #34 ?

casz:
Just checking if agent { label } + throttle with work same as agent none + throttle + node(label)

Are these two pipeline equal since throttle might run before reserving the executor?

```
pipeline {
    agent none
    stages {
        stage('test') {
            agent { label 'ios' }
            steps {
                throttle('ios-simulator') {
                    sh 'do stuff'
                }
            }
        }
    }
}
```

```
pipeline {
    agent none
    stages {
        stage('test') {
            agent none
            steps {
                throttle('ios-simulator') {
                    node('ios') {
                        sh 'do stuff'
                    }
                }
            }
        }
    }
}
```

casz:
Would it be possible to have agent label and throttle work so that we can have post steps that have file path context.

I would perhaps suggest this syntax but I know it requires work on the pipeline model definition instead.
This properly requires a JIRA issue, but just wondering if its at all possible :)

```
pipeline {
    agent none
    stages {
        stage('test') {
            agent {
                label 'ios'
                throttle 'ios-simulator'
            }
            steps {
                sh 'do stuff'
            }
        }
        post {
            always {
                junit '**/*.junit'
            }
        }
    }
}
```


This one below does not work:
```
pipeline {
    agent none
    stages {
        stage('test') {
            agent { label 'ios' }
            steps {
                throttle('ios-simulator') {
                    sh 'do stuff'
                }
            }
        }
        post {
            always {
                junit '**/*.junit'
            }
        }
    }
}
```

pom.xml
```diff
@@ -26,13 +26,13 @@ THE SOFTWARE.
...
     <artifactId>plugin</artifactId>
-    <version>2.6</version>
+    <version>2.22</version>
   </parent>
 
   <artifactId>throttle-concurrents</artifactId>
   <packaging>hpi</packaging>
   <name>Jenkins Throttle Concurrent Builds Plug-in</name>
-  <version>1.9.1-SNAPSHOT</version>
+  <version>2.0-SNAPSHOT</version>
```
oleg-nenashev: I would advice going through the beta- or -rc for this change. The potential impact is too high.

pom.xml
oleg-nenashev: Use `java.level`, Luke

pom.xml
```diff
@@ -105,29 +105,87 @@ THE SOFTWARE.
...
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-step-api</artifactId>
+            <version>2.7</version>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-durable-task-step</artifactId>
+            <version>2.8</version>
```
oleg-nenashev: Is there a need to have it as required dependency? It pulls a significant number of Pipeline plugins into (like Pipeline Support & Co)

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStepExecution.java
```diff
@@ -0,0 +1,80 @@
...
+
+public class ThrottleStepExecution extends StepExecution {
+    private final ThrottleStep step;
+
+    public ThrottleStepExecution(@Nonnull ThrottleStep step, StepContext context) {
+        super(context);
+        this.step = step;
+    }
+
+    public String getCategory() {
```
oleg-nenashev: Nonnull I'd guess

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStep.java
```diff
@@ -0,0 +1,66 @@
...
+public class ThrottleStep extends Step implements Serializable {
+    private String category;
+
+    @DataBoundConstructor
+    public ThrottleStep(@Nonnull String category) {
+        this.category = category;
+    }
+
+    @Nonnull
+    public String getCategory() {
```
oleg-nenashev: Only one category? I would expect an opportunity to define multiple categories like I can do for Job Properties

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStepExecution.java
```diff
@@ -0,0 +1,80 @@
...
+
+        getContext().newBodyInvoker()
+                .withCallback(new Callback(runId, flowNodeId, getCategory()))
+                .start();
+        return false;
+    }
+
+    @Override
+    public void stop(Throwable cause) throws Exception {
+
```
oleg-nenashev: Should it also call `removeThrottledPipelineForCategory` ?

src/main/resources/hudson/plugins/throttleconcurrents/ThrottleStep/config.jelly
```diff
@@ -0,0 +1,7 @@
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define"
+         xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form">
+    <f:entry title="${%Category}" field="category">
```
oleg-nenashev: I would expect multiple categories

src/test/java/hudson/plugins/throttleconcurrents/ThrottleConcurrentTest.java
```diff
@@ -25,6 +26,7 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+@Ignore("Depends on a newer version of Guava than can be used with Pipeline")
```
oleg-nenashev: It is a pretty important test suite. I do not feel that it is really required to have such declarative syntax here, so maybe it worth just rewriting it

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStepExecution.java
```diff
@@ -0,0 +1,80 @@
...
+        super(context);
+        this.step = step;
+    }
+
+    public String getCategory() {
+        return step.getCategory();
+    }
+
+    @Override
+    public boolean start() throws Exception {
```
oleg-nenashev: Maybe makes sense to contribute some information to the build log using the BuildListener from the context

src/main/resources/hudson/plugins/throttleconcurrents/ThrottleStep/config.jelly
```diff
@@ -0,0 +1,7 @@
...
+        <f:textbox/>
+    </f:entry>
```
oleg-nenashev: BTW, I am perfectly fine with dropping the support of local job throttling without categories

pom.xml
abayer: +1.

pom.xml
abayer: Yes - we need `ExecutorStepExecution.class` for checking against, etc.

pom.xml
abayer: Yessir. =)

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStepExecution.java
abayer: Yup.

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStepExecution.java
abayer: Good point.

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStepExecution.java
abayer: Apparently, from Jesse's earlier comments, no - `stop` doesn't have relevance here.

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStep.java
abayer: I'd rather keep it simple and wait to see what the demand is like for more complex.

src/main/resources/hudson/plugins/throttleconcurrents/ThrottleStep/config.jelly
abayer: See above.

src/test/java/hudson/plugins/throttleconcurrents/ThrottleConcurrentTest.java
abayer: I agree. I'll tackle that either here or in a follow-up PR.

abayer: Actually, on second thought, this is basically covered by `ThrottleStepTest`, isn't it?

abayer: I take that back - did some experimenting and there is something weird once you've got enough Pipeline tasks in queue. Working on it.

abayer: Aaaah, `queue.getPendingItems()`. Fixed.

src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
jglick: No, it is just `Serializable`.

abayer: Ah. Still not sure I want to actually save that large a chunk of data when I can just do the round-tripping...

abayer:
@oleg-nenashev What do you think now?

casz:
Been testing without issues running on 5 nodes trying to throttle executor for iOS simulator :)

oleg-nenashev:
will review it tomorrow

abayer:
@jglick Saw you added yourself as a reviewer again - want to review this still or can I merge?

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStep.java
ilker-demir: I would really like to be able to throttle multiple categories as this is our usual scenario for our freestyle jobs.

oleg-nenashev:
@abayer After some consideration and discussions with users at "CloudBees | Jenkins Automotive and Embedded Day", I think it makes sense to have support of multiple categories in the first iteration. Would you be available for a sync-up today?

abayer:
@oleg-nenashev At first glance, I can't see a way to do multiple categories as parameters to the step cleanly - i.e., we could do `throttle("category1,category2,category3")` or `throttle(["category1","category2","category3"])` but I don't like either one of those.

Wouldn't something like this work as is?
```groovy
throttle('category1') {
  throttle('category2') {
    // do stuff
  }
}
```

abayer:
Working on tests to make sure the above example works now.

oleg-nenashev:
@abayer 

> Wouldn't something like this work as is?

Such approach implies a risk of deadlock between runs If they acquire resources in a different order. I would not vote for recommending such approach to plugin users

abayer:
@oleg-nenashev Not sure I understand?

abayer:
Check that, think I hit the deadlock in testing. =)

pom.xml
jglick: Harmless, just stack traces. `workflow-support` itself should be bumped up.

src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
```diff
@@ -409,6 +505,77 @@ private int buildsOnExecutor(Task task, Executor exec) {
...
+        }
+
+        return 0;
+    }
+
+    private boolean isTaskThrottledPipeline(Task origTask, List<FlowNode> flowNodes) {
+        if (origTask instanceof PlaceholderTask) {
+            PlaceholderTask task = (PlaceholderTask)origTask;
+            try {
+                FlowNode firstThrottle = firstThrottleStartNode(task.getNode());
```
jglick: This really deserves a proper API so you can get rid of the `workflow-durable-task-step` dependency. It was never intended as an API plugin.

src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
```diff
@@ -409,6 +505,77 @@ private int buildsOnExecutor(Task task, Executor exec) {
...
+                if (enclosing != null &&
+                        enclosing instanceof BlockStartNode &&
+                        enclosing instanceof StepNode &&
+                        // There are two BlockStartNodes (aka StepStartNodes) for ThrottleStep, so make sure we get the
+                        // first one of those two, which will not have BodyInvocationAction.class on it.
+                        enclosing.getAction(BodyInvocationAction.class) == null) {
+                    // Check if this is a *different* throttling node.
+                    StepDescriptor desc = ((StepNode) enclosing).getDescriptor();
+                    if (desc != null && desc.getClass().equals(ThrottleStep.DescriptorImpl.class)) {
+                        return enclosing;
```
jglick: Oof. @svanoort?

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStep.java
```diff
@@ -0,0 +1,81 @@
...
+
+import javax.annotation.Nonnull;
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.StringTokenizer;
+
+public class ThrottleStep extends Step implements Serializable {
```
jglick: If you just have the one field, probably better to inline that into the execution state.

abayer:
@oleg-nenashev So are we good to go here? I defer to you for merging and releasing.

pom.xml
oleg-nenashev: You do not use this class from what I see

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStep.java
```diff
@@ -0,0 +1,81 @@
...
+        return categories;
+    }
+
+    @Nonnull
+    public List<String> getCategoriesList() {
+        List<String> catList = new ArrayList<>();
+
+        StringTokenizer tokenizer = new StringTokenizer(categories, ",");
+        while (tokenizer.hasMoreTokens()) {
+            catList.add(tokenizer.nextToken().trim());
```
oleg-nenashev: If I understand correctly, it is going to inject empty strings. Which may become a real problem if somebody uses something like `cat_a,${myvar},cat_c` expression

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStep.java
```diff
@@ -0,0 +1,81 @@
...
+        public boolean takesImplicitBlockArgument() {
+            return true;
+        }
+
+        @Override
+        public Set<? extends Class<?>> getRequiredContext() {
+            return Collections.singleton(TaskListener.class);
+        }
+
+        public FormValidation doCheckCategoryName(@QueryParameter String value) {
```
oleg-nenashev: Restricted ? not sure if RequirePost is mandatory

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStep.java
```diff
@@ -0,0 +1,81 @@
...
+    @Extension
+    public static final class DescriptorImpl extends StepDescriptor {
+        @Override
+        public String getFunctionName() {
+            return "throttle";
+        }
+
+        @Override
+        public String getDisplayName() {
+            return "Throttle execution of node blocks within this body";
```
oleg-nenashev: Nice2have: make it localizable

src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
```diff
@@ -349,6 +404,22 @@ private boolean isAnotherBuildWithSameParametersRunningOnNode(Node node, Queue.I
...
+            PlaceholderTask placeholderTask = (PlaceholderTask)task;
+            try {
+                FlowNode firstThrottle = firstThrottleStartNode(placeholderTask.getNode());
+                Run<?,?> r = placeholderTask.run();
+                if (firstThrottle != null && r != null) {
+                    return ThrottleJobProperty.getCategoriesForRunAndFlowNode(r.getExternalizableId(),
+                            firstThrottle.getId());
+                }
+            } catch (IOException | InterruptedException e) {
+                return new ArrayList<>();
```
oleg-nenashev: No logging? 🐜

src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
```diff
@@ -349,6 +404,22 @@ private boolean isAnotherBuildWithSameParametersRunningOnNode(Node node, Queue.I
...
+            }
+        }
+        return new ArrayList<>();
```
oleg-nenashev: Nice2have: assert : false

pom.xml
abayer: Ah, that may have gone away. Fixing.

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStep.java
abayer: Reasonable.

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStep.java
abayer: Not sure what you mean?

pom.xml
abayer: Nope, it's still there - specifically `ExecutorStepExecution.PlaceholderTask` in `ThrottleQueueTaskDispatcher`.

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStep.java
abayer: Good question. This doesn't actually expose any meaningful information - all you can do is tell whether there's a category with a given name, not anything about the category, etc. I think it's harmless, but defer to others.

src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
abayer: Will do.

src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
abayer: ? We're calling this method for all tasks, so an assert would cause problems.

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStep.java
oleg-nenashev: I mean strings like `throttle("cat_a,,cat_b")`, which may come to this method

abayer: Aaaaaah. Changing to discard empty.

abayer:
@oleg-nenashev Thanks for testing it out!

Snippet generator stuff definitely needs to be updated - I'll get that done.

> Multiple categories are being passed as string, not as a list/array. It complicates management of categories in the code
So this is an interesting question. I've found surprisingly few examples of steps taking lists, so I went conservative. I'll play around with this to see if I can get a good list or array approach.

> If the declared category does not exist, it is just ignored.
I wasn't sure what the best behavior would be in this case. I opted for a "first, do no harm" approach, but am fine with switching to fail.

> Behavior for duplicated definitions is not clear.
Huh. I literally never thought about that. So you're thinking logging something when there are duplicate categories?

> "Throttle Concurrent Builds" Job Property appears in the UI, but does nothing for both modes from what I see.
This isn't a change introduced by this PR - I think I did that a while back. It would, I think, apply to the full Pipeline execution, but yeah, it's pretty useless. I just worry that someone out there may be using it and would get screwed if we dropped it.

oleg-nenashev:
> I wasn't sure what the best behavior would be in this case. I opted for a "first, do no harm" approach, but am fine with switching to fail.

Yeah, this is my approach as well. I just expect less harm if the job fails than if it gets executed incorrectly. Better fail than ~undefined behavior~ sorry :)

> So you're thinking logging something when there are duplicate categories?

Or I can probably just add note to README in #47.

> This isn't a change introduced by this PR - I think I did that a while back. It would, I think, apply to the full Pipeline execution, but yeah, it's pretty useless. I just worry that someone out there may be using it and would get screwed if we dropped it.

I will check if it actually works. If yes, likely we need to polish UI a bit to prevent confusion of users. Not sure how exactly

abayer:
@oleg-nenashev Ok, I think I've responded to everything but the job property thing (which is a separate matter entirely).

oleg-nenashev:
@abayer After some consideration, I still feel bad about the TCB JobProperty UX. Would it be possible to add a disclaimer to its built-in help at least?

abayer:
@oleg-nenashev I...guess? I'd almost rather just make it not show up and say "screw it" to anyone who's for some reason somehow using the job property with a Pipeline currently.

abayer:
Also, I'm not the one who added that support in the first place, as far as I can tell. =)

oleg-nenashev:
@abayer 

> I'd almost rather just make it not show up and say "screw it" to anyone who's for some reason somehow using the job property with a Pipeline currently.

It is 2.0, so probably it is a valid option. The possible confusion is too high since the behavior is not documented at all. And the plugin explicitly says Pipeline is not supported. Just add incompatibleSince to pom.xml in such case

@casz @jenkinsci/code-reviewers If you use Global config for TCB, please let us know

oleg-nenashev:
The Pipeline support has been added by @dsavenko in https://github.com/jenkinsci/throttle-concurrent-builds-plugin/pull/32 . And yeah, both me and @abayer approved it. So I would rather vote for leaving as is and documenting

abayer:
@oleg-nenashev Then frankly, I think documenting it is a separate issue, isn't it?

oleg-nenashev:
@abayer 
> Then frankly, I think documenting it is a separate issue, isn't it?

 think some basic built-in docs should be included in this PR, because the new functionality leads to confusion with the existing one. A couple of sentences in JobProperty is enough. Regarding the full documentation, I will add it in https://github.com/jenkinsci/throttle-concurrent-builds-plugin/pull/47 . Does it sound as a deal?

abayer:
Deal.

src/main/java/hudson/plugins/throttleconcurrents/pipeline/ThrottleStep.java
```diff
@@ -80,6 +63,14 @@ public boolean takesImplicitBlockArgument() {
         public FormValidation doCheckCategoryName(@QueryParameter String value) {
             return ThrottleJobProperty.fetchDescriptor().doCheckCategoryName(value);
         }
+
+        public List<ThrottleJobProperty.ThrottleCategory> getCategories() {
+            return ThrottleJobProperty.fetchDescriptor().getCategories();
```
oleg-nenashev: Wrap to the unmodifyableList?

oleg-nenashev:
@reviewbybees done
I will test it tomorrow & merge it if everything is fine

oleg-nenashev:
All my comments have been addressed. I have created https://issues.jenkins-ci.org/browse/JENKINS-44373 for better diagnosability, but it is no a blocker for the merge. @abayer should I squash it?

PR #50, opened 2017-06-07
[JENKINS-44747] Limit the amount of time categoriesForPipeline can run

jglick:
[JENKINS-44747](https://issues.jenkins-ci.org/browse/JENKINS-44747)

Not sure if this is the right fix, but seems prudent to start here at least. Untested, since I do not know how to reproduce.

I suspect the root issue is that an `ExecutorPickle` (in the posted thread dump, called from `jenkins.util.Timer [#1]`) is trying to resume a `node` block from some build, but at the same time the dispatcher is attempting to decide whether to take another `node` block from the same build; until all the pickles are rehydrated successfully, the build will not resume, and `StepContext.get` will wait—hence a deadlock. (`ThrottleStepTest` does seem to check this kind of scenario, so perhaps it is timing-dependent if so.)

`ExecutorStepExecution.getNode` could perhaps check `!context.ready` and return `null` but this might be considered surprising. Anyway, modulo logging, this would result in the same behavior—an empty list of categories, thus allowing the block to proceed—so if there is a better fix it would presumably to have to stop relying on `FlowNode.id`. Perhaps you could key blocks by `PlaceholderTask.cookie` instead, though I am loath to expose this as an API.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -176,7 +176,6 @@ THE SOFTWARE.
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-support</artifactId>
             <version>2.13</version>
-            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
+++ b/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
@@ -27,6 +27,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
@@ -42,6 +43,7 @@ import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.graph.StepNode;
 import org.jenkinsci.plugins.workflow.graphanalysis.LinearBlockHoppingScanner;
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
+import org.jenkinsci.plugins.workflow.support.concurrent.Timeout;
 import org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution.PlaceholderTask;
 
 @Extension
@@ -407,7 +409,7 @@ public class ThrottleQueueTaskDispatcher extends QueueTaskDispatcher {
     private List<String> categoriesForPipeline(Task task) {
         if (task instanceof PlaceholderTask) {
             PlaceholderTask placeholderTask = (PlaceholderTask)task;
-            try {
+            try (Timeout t = Timeout.limit(100, TimeUnit.MILLISECONDS)) {
                 FlowNode firstThrottle = firstThrottleStartNode(placeholderTask.getNode());
                 Run<?,?> r = placeholderTask.run();
                 if (firstThrottle != null && r != null) {


pom.xml
```diff
@@ -176,7 +176,6 @@ THE SOFTWARE.
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-support</artifactId>
             <version>2.13</version>
-            <scope>test</scope>
```
oleg-nenashev: I am not a fan of it, it pulls in a whole bunch of non-API Pipeline plugins 🐜

jglick:
> From what I see this change actually allows to bypass the category check when the placeholder task is not initialized.

Yes—but in fact that was already happening, due to the call to `.run()`. So this patch should not make the behavior any less correct.

pom.xml
jglick: `workflow-support` was already a dependency, via `workflow-durable-task-step`. The POM formerly overrode the scope to be `test` (it would otherwise have defaulted to `compile`). Since I needed a facility introduced in 2.13, this sufficed.

jglick:
@oleg-nenashev we could do something like this

```diff
diff --git a/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java b/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
index 6d33e11..671179c 100644
--- a/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
+++ b/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
@@ -71,7 +71,13 @@ public class ThrottleQueueTaskDispatcher extends QueueTaskDispatcher {
     private CauseOfBlockage canTakeImpl(Node node, Task task) {
         final Jenkins jenkins = Jenkins.getActiveInstance();
         ThrottleJobProperty tjp = getThrottleJobProperty(task);
-        List<String> pipelineCategories = categoriesForPipeline(task);
+        List<String> pipelineCategories;
+        try {
+            pipelineCategories = categoriesForPipeline(task);
+        } catch (Exception x) {
+            LOGGER.log(Level.FINE, "Error getting categories for " + task, x);
+            return CauseOfBlockage.fromMessage(Messages._ThrottleQueueTaskDispatcher_build_not_ready());
+        }
 
         // Handle multi-configuration filters
         if (!shouldBeThrottled(task, tjp) && pipelineCategories.isEmpty()) {
@@ -167,7 +173,14 @@ public class ThrottleQueueTaskDispatcher extends QueueTaskDispatcher {
     // @Override on jenkins 4.127+ , but still compatible with 1.399
     public CauseOfBlockage canRun(Queue.Item item) {
         ThrottleJobProperty tjp = getThrottleJobProperty(item.task);
-        List<String> pipelineCategories = categoriesForPipeline(item.task);
+        List<String> pipelineCategories;
+        try {
+            pipelineCategories = categoriesForPipeline(item.task);
+        } catch (Exception x) {
+            LOGGER.log(Level.FINE, "Error getting categories for " + item.task, x);
+            // Should not be necessary to print more details, but if so, PlaceholderTask.runForDisplay would be appropriate.
+            return CauseOfBlockage.fromMessage(Messages._ThrottleQueueTaskDispatcher_build_not_ready());
+        }
 
         if (!pipelineCategories.isEmpty() || (tjp!=null && tjp.getThrottleEnabled())) {
             if (tjp != null && tjp.isLimitOneJobWithMatchingParams() && isAnotherBuildWithSameParametersRunningOnAnyNode(item)) {
@@ -407,7 +420,7 @@ public class ThrottleQueueTaskDispatcher extends QueueTaskDispatcher {
     }
 
     @Nonnull
-    private List<String> categoriesForPipeline(Task task) {
+    private List<String> categoriesForPipeline(Task task) throws IOException, InterruptedException {
         if (task instanceof PlaceholderTask) {
             PlaceholderTask placeholderTask = (PlaceholderTask)task;
             Run<?, ?> r = placeholderTask.run();
@@ -422,12 +435,10 @@ public class ThrottleQueueTaskDispatcher extends QueueTaskDispatcher {
                                 return categories;
                             }
                         }
-                    } catch (IOException | InterruptedException e) {
-                        LOGGER.log(Level.WARNING, "Error getting categories for pipeline {0}: {1}",
-                                new Object[] {task.getDisplayName(), e});
-                        return new ArrayList<>();
                     }
                 }
+            } else {
+                throw new IOException("build not ready"); // !context.ready
             }
         }
         return new ArrayList<>();
diff --git a/src/main/resources/hudson/plugins/throttleconcurrents/Messages.properties b/src/main/resources/hudson/plugins/throttleconcurrents/Messages.properties
index 50ba33d..74bcb69 100644
--- a/src/main/resources/hudson/plugins/throttleconcurrents/Messages.properties
+++ b/src/main/resources/hudson/plugins/throttleconcurrents/Messages.properties
@@ -2,6 +2,7 @@ ThrottleQueueTaskDispatcher.MaxCapacityOnNode=Already running {0} builds on node
 ThrottleQueueTaskDispatcher.MaxCapacityTotal=Already running {0} builds across all nodes
 ThrottleQueueTaskDispatcher.BuildPending=A build is pending launch
 ThrottleQueueTaskDispatcher.OnlyOneWithMatchingParameters=A build with matching parameters is already running
+ThrottleQueueTaskDispatcher.build_not_ready=Build is not ready to resume
 
 ThrottleMatrixProjectOptions.DisplayName=Additional options for Matrix projects
 
```

but I suspect this will just reintroduce the bug in a different form: so long as the build is not fully resumed (which requires all `node` blocks to be actually executing, thus not blocked by the dispatcher), `!context.ready` and thus the dispatcher will block. We would not be holding the `Queue` lock, which is good, but the build would hang. This all goes back to the fundamental design mistake I have discussed several times with @kohsuke, that `StepExecution`s should have been responsible for resuming their virtual threads on their own timeline, rather than relying on the `Pickle` system.

Again, to really solve this you would need to use a different key. Which would be a great enhancement, if someone can figure out how to do it, and especially how to write a test reproducing the situation. In the meantime, this patch ought to solve the critical `Queue` deadlock, which probably affects even jobs not using this plugin at all.

oleg-nenashev:
Yes, whatever hotfix is useful. But effectively we need to create follow-up
bugs and to add the restart case to the plugin's Errata. Smells bad, fine
as am interim solution

On Jun 7, 2017 7:52 PM, "Jesse Glick" <notifications@github.com> wrote:

@oleg-nenashev <https://github.com/oleg-nenashev> we could do something
like this

diff --git a/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
b/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
index 6d33e11..671179c 100644---
a/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java+++
b/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java@@
-71,7 +71,13 @@ public class ThrottleQueueTaskDispatcher extends
QueueTaskDispatcher {
     private CauseOfBlockage canTakeImpl(Node node, Task task) {
         final Jenkins jenkins = Jenkins.getActiveInstance();
         ThrottleJobProperty tjp = getThrottleJobProperty(task);-
  List<String> pipelineCategories = categoriesForPipeline(task);+
  List<String> pipelineCategories;+        try {+
pipelineCategories = categoriesForPipeline(task);+        } catch
(Exception x) {+            LOGGER.log(Level.FINE, "Error getting
categories for " + task, x);+            return
CauseOfBlockage.fromMessage(Messages._ThrottleQueueTaskDispatcher_build_not_ready());+
       }

         // Handle multi-configuration filters
         if (!shouldBeThrottled(task, tjp) &&
pipelineCategories.isEmpty()) {@@ -167,7 +173,14 @@ public class
ThrottleQueueTaskDispatcher extends QueueTaskDispatcher {
     // @Override on jenkins 4.127+ , but still compatible with 1.399
     public CauseOfBlockage canRun(Queue.Item item) {
         ThrottleJobProperty tjp = getThrottleJobProperty(item.task);-
       List<String> pipelineCategories =
categoriesForPipeline(item.task);+        List<String>
pipelineCategories;+        try {+            pipelineCategories =
categoriesForPipeline(item.task);+        } catch (Exception x) {+
       LOGGER.log(Level.FINE, "Error getting categories for " +
item.task, x);+            // Should not be necessary to print more
details, but if so, PlaceholderTask.runForDisplay would be
appropriate.+            return
CauseOfBlockage.fromMessage(Messages._ThrottleQueueTaskDispatcher_build_not_ready());+
       }

         if (!pipelineCategories.isEmpty() || (tjp!=null &&
tjp.getThrottleEnabled())) {
             if (tjp != null && tjp.isLimitOneJobWithMatchingParams()
&& isAnotherBuildWithSameParametersRunningOnAnyNode(item)) {@@ -407,7
+420,7 @@ public class ThrottleQueueTaskDispatcher extends
QueueTaskDispatcher {
     }

     @Nonnull-    private List<String> categoriesForPipeline(Task
task) {+    private List<String> categoriesForPipeline(Task task)
throws IOException, InterruptedException {
         if (task instanceof PlaceholderTask) {
             PlaceholderTask placeholderTask = (PlaceholderTask)task;
             Run<?, ?> r = placeholderTask.run();@@ -422,12 +435,10 @@
public class ThrottleQueueTaskDispatcher extends QueueTaskDispatcher {
                                 return categories;
                             }
                         }-                    } catch (IOException |
InterruptedException e) {-
LOGGER.log(Level.WARNING, "Error getting categories for pipeline {0}:
{1}",-                                new Object[]
{task.getDisplayName(), e});-                        return new
ArrayList<>();
                     }
                 }+            } else {+                throw new
IOException("build not ready"); // !context.ready
             }
         }
         return new ArrayList<>();diff --git
a/src/main/resources/hudson/plugins/throttleconcurrents/Messages.properties
b/src/main/resources/hudson/plugins/throttleconcurrents/Messages.properties
index 50ba33d..74bcb69 100644---
a/src/main/resources/hudson/plugins/throttleconcurrents/Messages.properties+++
b/src/main/resources/hudson/plugins/throttleconcurrents/Messages.properties@@
-2,6 +2,7 @@ ThrottleQueueTaskDispatcher.MaxCapacityOnNode=Already
running {0} builds on node
 ThrottleQueueTaskDispatcher.MaxCapacityTotal=Already running {0}
builds across all nodes
 ThrottleQueueTaskDispatcher.BuildPending=A build is pending launch
 ThrottleQueueTaskDispatcher.OnlyOneWithMatchingParameters=A build
with matching parameters is already
running+ThrottleQueueTaskDispatcher.build_not_ready=Build is not ready
to resume

 ThrottleMatrixProjectOptions.DisplayName=Additional options for Matrix projects


but I suspect this will just reintroduce the bug in a different form: so
long as the build is not fully resumed (which requires all node blocks to
be actually executing, thus not blocked by the dispatcher), !context.ready
and thus the dispatcher will block. We would not be holding the Queue lock,
which is good, but the build would hang. This all goes back to the
fundamental design mistake I have discussed several times with @kohsuke
<https://github.com/kohsuke>, that StepExecutions should have been
responsible for resuming their virtual threads on their own timeline,
rather than relying on the Pickle system.

Again, to really solve this you would need to use a different key. Which
would be a great enhancement, if someone can figure out how to do it, and
especially how to write a test reproducing the situation. In the meantime,
this patch ought to solve the critical Queue deadlock, which probably
affects even jobs not using this plugin at all.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/jenkinsci/throttle-concurrent-builds-plugin/pull/50#issuecomment-306873597>,
or mute the thread

abayer:
Since @oleg-nenashev approved over on #51, I'm gonna merge this and then merge that so that the PR history is more clear.

jglick:
@abayer you can just merge a subsuming PR. If you do not squash or rebase, GitHub will know what to do.

PR #53, opened 2017-08-02
POM scm url was wrong.

abayer:
This broke links like on https://jenkins.io/doc/developer/extensions/jenkins-core/#jobproperty

cc @reviewbybees


abayer:
@oleg-nenashev Won't work here due to my stupidity way back in the day - `artifactId` is `throttle-concurrents`.

PR #56, opened 2018-11-18
[JENKINS-54578] Cannot delete last category

aeoris:
It was impossible to delete all categories (neither at once nor one by
one), since when not receiving any categories, the last-saved categories
took precedence. This has been solved by initializing the categories
property when not receiving the categories from the configuration form.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -170,16 +170,19 @@ THE SOFTWARE.
             <groupId>org.hamcrest</groupId>
             <artifactId>hamcrest-library</artifactId>
             <version>1.2.1</version>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.hamcrest</groupId>
             <artifactId>hamcrest-core</artifactId>
             <version>1.2.1</version>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.hamcrest</groupId>
             <artifactId>hamcrest-all</artifactId>
             <version>1.3</version>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -166,6 +166,21 @@ THE SOFTWARE.
             <version>2.3</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-library</artifactId>
+            <version>1.2.1</version>
```
oleg-nenashev: If needed should be a test dependency.
But generally it should be already coming from parent POM.

aeoris: Done! The first build failed, and was solved by adding those dependencies as suggested [in the wiki](https://wiki.jenkins.io/display/JENKINS/Unit+Test#UnitTest-pom.xml) (which I just fixed to include the scope as well).

I will read about Jelly forms to open a different PR updating them if that's OK.

basil:
Thanks for your contribution, @aeoris!

I just merged your PR with `master` and made a few edits:

- Added a defensive null check in `hudson.plugins.throttleconcurrents.ThrottleJobProperty.DescriptorImpl#getCategoryByName`.
- Rewrote the test to use `org.jvnet.hudson.test.JenkinsRule#createWebClient` rather than mocking. This results in more realistic testing against an actual Jenkins instance.

This should be a part of the forthcoming Throttle Concurrent Builds 2.0.2 release. Thanks again!

basil:
Released in [Throttle Concurrent Builds 2.0.2](https://github.com/jenkinsci/throttle-concurrent-builds-plugin/releases/tag/throttle-concurrents-2.0.2).

PR #57, opened 2019-02-13
Use FlowExecutionList to calculate the number of running pipeline jobs

basil:
### Overview

See [JENKINS-45571](https://issues.jenkins-ci.org/browse/JENKINS-45571) and [JENKINS-53223](https://issues.jenkins-ci.org/browse/JENKINS-53223).

### Problem

Pipeline sometimes inscrutably leaks flyweight executors. When this happens, it breaks the Throttle Concurrent Builds plugin, which counts flyweight executors when determining how many builds of a job are running. Not only does this result in builds being erroneously throttled, but it also chews up CPU on the Jenkins master in stacks like this:

```
"jenkins.util.Timer [#4]" #77 daemon prio=5 os_prio=0 tid=0x00007f50a800e800 nid=0x80d runnable [0x00007f504788a000]
   java.lang.Thread.State: RUNNABLE
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireShared(AbstractQueuedSynchronizer.java:1282)
        at java.util.concurrent.locks.ReentrantReadWriteLock$ReadLock.lock(ReentrantReadWriteLock.java:727)
        at hudson.model.Executor.getCurrentExecutable(Executor.java:514)
        at hudson.plugins.throttleconcurrents.ThrottleQueueTaskDispatcher.buildsOnExecutor(ThrottleQueueTaskDispatcher.java:511)
        at hudson.plugins.throttleconcurrents.ThrottleQueueTaskDispatcher.buildsOfProjectOnNode(ThrottleQueueTaskDispatcher.java:488)
        at hudson.plugins.throttleconcurrents.ThrottleQueueTaskDispatcher.buildsOfProjectOnAllNodes(ThrottleQueueTaskDispatcher.java:501)
        at hudson.plugins.throttleconcurrents.ThrottleQueueTaskDispatcher.throttleCheckForCategoriesAllNodes(ThrottleQueueTaskDispatcher.java:281)
        at hudson.plugins.throttleconcurrents.ThrottleQueueTaskDispatcher.canRunImpl(ThrottleQueueTaskDispatcher.java:253)
        at hudson.plugins.throttleconcurrents.ThrottleQueueTaskDispatcher.canRun(ThrottleQueueTaskDispatcher.java:218)
        at hudson.plugins.throttleconcurrents.ThrottleQueueTaskDispatcher.canRun(ThrottleQueueTaskDispatcher.java:176)
        at hudson.model.Queue.getCauseOfBlockageForItem(Queue.java:1197)
        at hudson.model.Queue.maintain(Queue.java:1522)
[...]
```

Last night, I found over 200 threads in this state on one of my Jenkins masters. They were using so much CPU that the entire master became unresponsive.

### Solution

As @dwnusbaum wrote in JENKINS-53223, "Without a consistent and simple reproduction case, … we are [not] going to be able to make any progress on fixing the root cause of flyweight executors leaking anytime soon, and since flyweight executors do not use a thread or other resources it doesn't really matter if there are a bunch of them in the system except for the fact that Throttle Concurrent Builds uses them to determine what is currently running." Therefore, I am attempting to modify the Throttle Concurrent Builds plugin to directly examine running builds instead of using executors as a proxy, as @jglick suggested in JENKINS-45571.

### DIFF
--- a/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
+++ b/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
@@ -38,6 +38,8 @@ import org.acegisecurity.context.SecurityContextHolder;
 
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.workflow.actions.BodyInvocationAction;
+import org.jenkinsci.plugins.workflow.flow.FlowExecution;
+import org.jenkinsci.plugins.workflow.flow.FlowExecutionList;
 import org.jenkinsci.plugins.workflow.graph.BlockStartNode;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.graph.StepNode;
@@ -476,7 +478,44 @@ public class ThrottleQueueTaskDispatcher extends QueueTaskDispatcher {
             return 0;
         }
 
+        // Note that this counts flyweight executors in its calculation, which may be a problem if flyweight executors
+        // are being leaked by other plugins.
+        return buildsOfProjectOnNodeImpl(node, task);
+    }
+
+    private int buildsOfProjectOnAllNodes(Task task) {
+        if (!shouldBeThrottled(task, getThrottleJobProperty(task))) {
+            return 0;
+        }
+
+        // Note that we can't use WorkflowJob.class because it is not on this plugin's classpath.
+        if (task.getClass().getName().equals("org.jenkinsci.plugins.workflow.job.WorkflowJob")) {
+            return buildsOfPipelineJob(task);
+        } else {
+            return buildsOfProjectOnAllNodesImpl(task);
+        }
+    }
+
+    private int buildsOfPipelineJob(Task task) {
         int runCount = 0;
+
+        for (FlowExecution flowExecution : FlowExecutionList.get()) {
+            try {
+                final Queue.Executable executable = flowExecution.getOwner().getExecutable();
+                if (executable != null && task.equals(executable.getParent())) {
+                    runCount++;
+                }
+            } catch (IOException e) {
+                LOGGER.log(Level.WARNING, "Error getting number of builds for pipeline {0}: {1}", new Object[] {task.getDisplayName(), e});
+            }
+        }
+
+        return runCount;
+    }
+
+    private int buildsOfProjectOnNodeImpl(Node node, Task task) {
+        int runCount = 0;
+
         LOGGER.log(Level.FINE, "Checking for builds of {0} on node {1}", new Object[] {task.getName(), node.getDisplayName()});
 
         // I think this'll be more reliable than job.getBuilds(), which seemed to not always get
@@ -496,12 +535,12 @@ public class ThrottleQueueTaskDispatcher extends QueueTaskDispatcher {
         return runCount;
     }
 
-    private int buildsOfProjectOnAllNodes(Task task) {
+    private int buildsOfProjectOnAllNodesImpl(Task task) {
         final Jenkins jenkins = Jenkins.getActiveInstance();
         int totalRunCount = buildsOfProjectOnNode(jenkins, task);
 
         for (Node node : jenkins.getNodes()) {
-            totalRunCount += buildsOfProjectOnNode(node, task);
+            totalRunCount += buildsOfProjectOnNodeImpl(node, task);
         }
         return totalRunCount;
     }
--- a/src/test/java/hudson/plugins/throttleconcurrents/ThrottleIntegrationTest.java
+++ b/src/test/java/hudson/plugins/throttleconcurrents/ThrottleIntegrationTest.java
@@ -118,7 +118,7 @@ public class ThrottleIntegrationTest extends HudsonTestCase {
         assertEquals(2, waterMark.getExecutorWaterMark());
     }
     
-    public void testThrottlingWithCategory() throws Exception {
+    public void testThrottlingWithCategoryPerNode() throws Exception {
         setupSlave();
         setupSecurity();
         final String category = "category";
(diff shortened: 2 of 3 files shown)


src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
```diff
@@ -496,14 +500,46 @@ private int buildsOfProjectOnNode(Node node, Task task) {
...
+        Job job = taskToJob(task);
+        if (job != null) {
+            RunList<?> runningBuilds = job.getBuilds().filter(new Predicate<Run>() {
+                @Override
+                public boolean apply(Run run) {
+                    return run.isBuilding();
+                }
+            });
+
+            int runCount = Iterators.size(runningBuilds.iterator());
```
dwnusbaum: nit: perhaps better to gate this call behind `if (LOGGER.isLoggable(Level.FINE)` to avoid iterating through all builds twice.

src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
```diff
@@ -496,14 +500,46 @@ private int buildsOfProjectOnNode(Node node, Task task) {
...
+        if (!shouldBeThrottled(task, getThrottleJobProperty(task))) {
+            return 0;
+        }
 
-        for (Node node : jenkins.getNodes()) {
-            totalRunCount += buildsOfProjectOnNode(node, task);
+        // Prefer Job#getBuilds() to counting executors because the former doesn't count flyweight executors.
+        Job job = taskToJob(task);
+        if (job != null) {
+            RunList<?> runningBuilds = job.getBuilds().filter(new Predicate<Run>() {
```
dwnusbaum: I think this will cause all builds for the job to be loaded from disk, which seems like it could be a bad idea for performance if you have a lot of build history. You could use `FlowExecutionList` to get only running Pipelines and filter that down to the job in question, which has the benefit that all of those builds are guaranteed to already be loaded since they are running. Maybe even keep the old code path for non-Pipeline builds?

basil:
> [PlaceholderTask](https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/ae34b63554b8edca6b51e28d7a1936af733af431/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L262) is used by Pipeline `node` steps when they are resuming and waiting for their agent to become available. That code is probably needed to correctly count resuming Pipelines.

Since we last talked, I dug deeper into the various use cases for this plugin and answered my own question, wrote a few new test cases, and discovered some interesting results about how this plugin works when Jenkins is restarted.

There appear to be two main tests for this plugin: `ThrottleIntegrationTest` for Freestyle jobs and `ThrottleStepTest` for Pipeline jobs.

- `ThrottleIntegrationTest` contains a test for jobs that throttle the number of concurrent builds per node for a category; however, there is no test for jobs that throttle the total number of concurrent builds for a category. I added such a test.

- `ThrottleStepTest` contains a test for node-level throttling with a `throttle` step; however, there is no test for job-level throttling. I added such a test.

While executing the various tests in this class and stepping through the plugin code, I was able to answer my own question about when `PlaceholderTask` needs to be handled. It turns out that all the methods that use `PlaceholderTask` in the existing codebase are for codepaths that perform node-level throttling with a `throttle` step (which only applies to Pipeline jobs), and these methods need to deal with `PlaceholderTask` even when Jenkins is not restarted. The methods that are used for codepaths that perform job-level throttling (which apply to both Freestyle and Pipeline jobs) are never called with a `PlaceholderTask` as input even when Jenkins _is_ restarted. I verified the latter by writing a new test that restarts Jenkins with job-level throttling in place and ensured that in the job-level throttling codepaths the `task` argument was never a `PlaceholderTask`.

Now here is where things get interesting. All of my new tests passed against the existing codebase except for the test that restarts Jenkins. When I started debugging it, I found that the job-level throttle was not actually kicking in after Jenkins is restarted. I looked into it further and found that [this is a known bug in the existing codebase](https://issues.jenkins-ci.org/browse/JENKINS-44756). However, as you will see below, I was able to fix this bug as part of the changes you suggested in your last review.

> I guess one of the catches here is that `buildsOfProjectOnNode` is still used in some cases, and it's not clear how that method could be implemented without looking at executors, so that might be a problem depending on how it is used and the calling methods might need to be adjusted.

Yes, you are absolutely right that this may be a problem. There were previously three ways for pipeline jobs to go through the problematic code path that iterates through flyweight executors: the job-level throttling code path (which calls `buildsOfProjectOnAllNodes`, which calls `buildsOfProjectOnNode`) and two other callers that call `buildsOfProjectOnNode` directly. My change mitigates the problem significantly by addressing the job-level throttling code path (i.e., `buildsOfProjectOnAllNodes`), but it doesn't completely eliminate the problem since I haven't addressed the remaining two callers of `buildsOfProjectOnNode`. I couldn't figure out how to deal with these. I left a comment drawing attention to this, but I don't think I'll be able to fix these two callers without further design advice. I believe my change is a strict improvement over the current status quo and is still worth merging as-is.

I split up my change into two commits:

1. The first commit increases test coverage for the existing code base. These new tests pass against the existing code base. If you'd like, I could split these into a separate PR that we can merge first before continuing work on this PR.
2. The second commit implements the suggestions from your last review. These suggestions incidentally fix the restart bug, so in this commit I also add a new test that restarts Jenkins. This test only works with my fixes to the code under test.

src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
basil: After your last round of feedback, this code no longer exists, so this comment no longer applies.

src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
basil: Great catch! I've done exactly as you suggested, including keeping the old code path for non-Pipeline builds.

jglick:
> The methods that are used for codepaths that perform job-level throttling

Why would there be any such methods at all? This does not make sense to me. You could have ten thousands builds all waiting inside

```groovy
echo 'start'
input 'OK?' // ← here
echo 'end'
```

and consuming zero resources other than holding a little bit of heap for build metadata.

basil:
> Why would there be any such methods at all? This does not make sense to me.

Legacy interfaces often don't make much sense, but it is still important to keep them working in order to avoid breaking legacy consumers. I didn't create the pipeline job-level throttling support in this plugin, but our (and possibly others') jobs depend on it; therefore, it makes sense to me that it should be fixed to work properly in recent versions of Pipeline that leak flyweight executors.

basil:
This change has been running on our production Jenkins server for the past month without issue.

basil:
Released in [Throttle Concurrent Builds 2.0.2](https://github.com/jenkinsci/throttle-concurrent-builds-plugin/releases/tag/throttle-concurrents-2.0.2).

PR #38, opened 2016-02-22
Fixing throttle based on parameter examination (Replaces PR#33)

balldarrens:
Please review and comment.

@oleg-nenashev 
@jmozmoz
@abayer
@ikedam

PR #59, opened 2019-06-28
Fix same params detection (documented "comma" does not work)

jimklimov:
So far posting for review and sanity check.

We have an issue that several jobs are configured with a comma-separated list of parameters to check against for throttling (e.g. that we do not run tests against the same host and port in parallel), and expect that tests with other combinations of parameters would run and not stay in queue. We see this expected behavior in jobs that do not specify a list of parameters, and I think in those that specify just one.

So far we see instead that comparison is done to an empty array when we specify a list of parameters, as implemented in #38:
````
Jun 28, 2019 5:52:44 PM FINE hudson.plugins.throttleconcurrents.ThrottleQueueTaskDispatcher isAnotherBuildWithSameParametersRunningOnNode

build (hudson.model.queue.WorkUnit@5e3c8658[work=selenium]) with identical parameters ([]) is already running.
````
and so only one instance of a job is running and others (tests against different hosts) are queued.

This PR starts out as a bit of documentation and instrumentation to understand the problem, and I hope would end up with a solution too :)

### DIFF
--- a/src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
+++ b/src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
@@ -219,11 +219,11 @@ public class ThrottleJobProperty extends JobProperty<Job<?,?>> {
     }
 
     /**
-     * Set the paramsToCompare list of Strings from the user-input
-     * token-separated String.
+     * @param paramsToUseForLimit   Set the paramsToCompare list
+     * of Strings from the user-input token-separated String.
      *
-     * @return None, but the this.paramsToCompare will be a valid
-     * populated or empty array (not a null) after this call.
+     * There is no return, but the this.paramsToCompare will be a
+     * valid non-null populated or empty array after this call.
      */
     public void setParamsToCompare(String paramsToUseForLimit) {
         // If there is any contents, tell GC that it is reapable
--- a/src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
+++ b/src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
@@ -222,7 +222,7 @@ public class ThrottleJobProperty extends JobProperty<Job<?,?>> {
         if (paramsToCompare == null) {
             if ((paramsToUseForLimit != null)) {
                 if ((paramsToUseForLimit.length() > 0)) {
-                    paramsToCompare = Arrays.asList(paramsToUseForLimit.split(",\\s*|\\s+"));
+                    paramsToCompare = Arrays.asList(ArrayUtils.nullToEmpty(paramsToUseForLimit.split(",\\s*|\\s+")));
                 }
                 else {
                     paramsToCompare = new ArrayList<String>();
--- a/src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
+++ b/src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
@@ -51,10 +51,6 @@ public class ThrottleJobProperty extends JobProperty<Job<?,?>> {
     // Replaced by categories, to support, well, multiple categories per job (starting from 1.3)
     @Deprecated transient String category;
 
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by jimklimov** [comment]
So locally got to the root cause, beside the other clean-ups and instrumentations added above that were quite helpful for "in-vivo" research of issues with the plugin, so I'd rather submit them also.

The problem in the end was that the comma-aware setter (or rather getter that can set) was not really called, because a similar but different logic was in the constructor, which just used `split()` and so whitespaces, contrary to the doc.

I'll try to upload the working changeset here shortly. It may take a while (and a lot of self-convincing) to proceed on a maybe more proper fix and/or finding a way to unit-test this.

In any case, peer review of the practical change can commence soon :)

**Comment by jimklimov** [comment]
With the code and logic now testable, I proclaim this fully ready for review and eventual merge (aka "done on my side") ;)

File: src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java [hunk 1]
```diff
@@ -218,20 +218,54 @@ public boolean isThrottleMatrixConfigurations() {
...
+                list.removeAll(Arrays.asList("", " ", "\n", "\t", ","));
+                this.paramsToCompare = list;
             }
             else {
-                paramsToCompare = new ArrayList<String>();
+                this.paramsToCompare = new ArrayList<String>();
             }
         }
+        else {
+            this.paramsToCompare = new ArrayList<String>();
```
olivergondza: Might be worth using `Collections.emptyList()` while at it.

olivergondza:
I do not see anything obviously wrong here, but I known nothing about internal workings of the plugin.

src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
jimklimov: I thought of it while googling for modern patterns, but saw some value in keeping the minimum dependencies needed to build this (adding as few import's as I can).

Would the other solutions be more efficient for example? This is a moderately hot codepath I suppose (executed whenever a job might pass from queued to building).

olivergondza: It does not appear to be on a hot code path to me. Though still, `Collections` is part of JDK so the only dependency in play is adding the import line.

jimklimov: Oh, ok. I thought it was part of apache tooling.

jimklimov: I looked at it and am inclined to leave as is, to keep the code styled the same way and the type-casting clearer in the different conditional cases we have to set up the list contents here.

basil:
Thanks for your contribution, @jimklimov!

>This PR starts out as a bit of documentation and instrumentation to understand the problem, and I hope would end up with a solution too :)

Could you please break out the documentation and instrumentation into a separate PR? I am the new maintainer of this plugin and am preparing to release Throttle Concurrent Builds 2.0.2. Throttle Concurrent Builds 2.0.1 was released in June 2017, and Throttle Concurrent Builds 2.0.2 will be a minimal bug fix release primarily intended to deliver #57 and #59. However, your documentation and instrumentation changes seem low-risk, and as long as I can prove (through code coverage) that they are exercised by an existing test, I think I'd be wiling to include them in Throttle Concurrent Builds 2.0.2. The rest of the changes in this PR will have to wait until the release after that, though.

jimklimov:
Hello, sorry for a delay - I'm in the midst of migrating to a new computer and the usual end-of-year hassle. I'll try to split the PR as you asked, but not sure when I'd get to it... hopefully would before the end of year...

jimklimov:
This PR was split into three blocks of different complexity and intrusiveness, as asked, and another PR #78 was just posted to reapply these three back over current master branch in proper order and with conflict resolution to achieve same codebase state as in this PR. In essence, now #59 and #78 are duplicates at this time.

PR #75, opened 2020-02-10
Update docs/whitespace

jimklimov:
As @basil asked, here is a least problematic subset of #59 that adds comments in the code (about what happens in the logic), a bit of formatting fix to accomodate other changes from that PR, and a text in Jelly to help users separate the parameter names to check not only with commas but also whitespaces.

The actual fix from using whitespace de-facto vs./along-with commas as documented will be the focus of other PRs split from #59.

### DIFF
--- a/src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
+++ b/src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
@@ -65,7 +65,7 @@ public class ThrottleJobProperty extends JobProperty<Job<?,?>> {
     // The paramsToUseForLimit is assigned by end-user configuration and
     // is generally a string with names of build arguments to consider,
     // and is empty, or has one arg name, or a token-separated list of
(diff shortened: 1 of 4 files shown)


src/main/java/hudson/plugins/throttleconcurrents/ThrottleJobProperty.java
```diff
@@ -62,6 +62,12 @@
     private transient boolean throttleConfiguration;
     private @CheckForNull ThrottleMatrixProjectOptions matrixOptions;
 
+    // The paramsToUseForLimit is assigned by end-user configuration and
+    // is generally a string with names of build arguments to consider,
+    // and is empty, or has one arg name, or a token-separated list of
+    // such names (see PARAMS_LIMIT_SEPARATOR) below.
```
basil: The `PARAMS_LIMIT_SEPARATOR` field from #59 isn't being introduced yet in this PR, so can we remove the reference from the comment?

jimklimov: ack

PR #76, opened 2020-02-10
Params comparison instrumentation

jimklimov:
As @basil asked, here is another presumably not problematic subset of #59 that adds logging instrumentation in the code going through build parameter filtering. Due to the split of PRs, there may be a little formatting conflict vs. #75 that is another part of that original wad of code, and vs. subsequent PR with actual attempt at the fix.


jimklimov:
Updated for minor conflicts after #75 was merged

basil:
@jimklimov Is there still a strong motivation to finish and merge this PR? I'm not sure if this is still useful or if it has outlived its purpose after it was used to successfully debug previous issues.

If it is still needed, I'm concerned about the cost of the string concatenation for all the fine log statements. This is a hot code path in Jenkins, so any extra overhead here is a bit of a risk. The traditional best practice for this in Java is to guard such string concatenation with e.g. `if (logger.isDebugEnabled()) {`, but newer versions of Java allow for lazy computation of the string by using lambdas. [This article](https://garygregory.wordpress.com/2015/09/16/a-gentle-introduction-to-the-log4j-api-and-lambda-basics/) explains both approaches.

PR #77, opened 2020-02-10
Fix same params detection (change and test)

jimklimov:
As @basil asked, here is the potentially problematic subset of #59 that changes the logic of production code but also adds tests for that  :)

See also #75 and #76.

Note that due to the split from original PR, there may be minor merge-conflicts vs other pieces.


jimklimov:
Updated for minor conflicts after #75 was merged

basil:
Thanks for the PR! I refiled this PR as #103 and #104 with some additional changes.

basil:
Released in [2.1](https://github.com/jenkinsci/throttle-concurrent-builds-plugin/releases/tag/throttle-concurrents-2.1).

PR #95, opened 2020-12-03
ThrottleJobProperty/global.jelly: replace "table" by "div"

jimklimov:
Supposedly this factors into the table-to-div migration effort.

CC @timja


jimklimov:
Hm... no clue what in this PR can break the tests :\ Master was built a month ago and seems green...

PR #96, opened 2020-12-07
ThrottleJobProperty/global.jelly: replace "table" by "div"

jimklimov:
Fixes UI behavior for Jenkins 2.264+

CC @timja


src/main/resources/hudson/plugins/throttleconcurrents/ThrottleJobProperty/global.jelly
```diff
@@ -4,7 +4,7 @@
   <f:section title="Throttle Concurrent Builds">
     <f:entry title="Multi-Project Throttle Categories" field="categories">
       <f:repeatable field="categories" add="${%Add Category}" minimum="0">
-        <table width="100%">
+        <div style="width:100%">
```
timja, line 7: This needs either the minimum jenkins version increased or the compatibility code added:
https://www.jenkins.io/doc/developer/views/table-to-div-migration/#maintaining-support-for-tables-and-div-layout

basil:
Of [the users who have upgraded to the latest release](https://stats.jenkins.io/pluginversions/throttle-concurrents.html) (which are themselves 30% of the install base), only 24% are using a weekly Jenkins release newer than the most recent LTS. Rather than forcing the remaining 76% to choose between switching to a weekly Jenkins release newer than the most recent LTS or not receiving plugin updates, I think I would prefer to adopt the compatibility code and wait to bump the minimum Jenkins version until after the relevant upstream change has made it into at least one LTS release.

jimklimov:
That stance makes sense, but I believe many plugins had the compatibility
code (blockWrapper jelly) during the year until 2.264 and then dropped it
already in favor of explicit div tags.

Would it help to bump required core to just 2.264?

basil:
> Would it help to bump required core to just 2.264?

I don't think it would make much of a difference because 2.264 is still a weekly release newer than the latest LTS. The concept of a transition period (where the plugin supports both old and new cores) followed by a hard cutover (where the plugin only supports new cores) makes sense to me. Is there a particular reason why we would want to dispense with the transition period and go straight to the hard cutover with this plugin? In the presence of such a reason I could be convinced to go with this plan, but in the absence of such a reason I think we should have a transition period for the benefit of the majority of users.

jimklimov:
No particular preference or reason on my side, just following precedent
seen in other plugins and suggestions from @timja.

timja:
Most plugins adapt the compatibility code, unless it’s not straightforward

(Non straightforward is heavy use of JavaScript, forked core widgets / very custom widgets normally)

jimklimov:
Oh ok, I thought I saw some drop the shim already. Certainly saw comments
like "this should be gone after 2.264".

oleg-nenashev (commented):
Let's rather add the compatibility layer, should work OOTB here

jimklimov:
Makes sense, but I'm afraid I can't commit to completing this ASAPish soon - family's got plans away from computers. So if anyone takes over to deliver faster than I get to get back to this, I won't mind ;)

jimklimov:
I guess I was under impression from https://github.com/jenkinsci/multiple-scms-plugin/pull/25/files that is linked from that conversion doc page. This PR outright drops the shim file - but now I've got where to pick it up :)

jimklimov:
Hopefully done, to the best understanding of monkey copy-paste ;)

src/main/resources/lib/throttleconcurrents/blockWrapper.jelly
```diff
@@ -0,0 +1,16 @@
+<?jelly escape-by-default='true'?>
```
timja, line 1: needs the 'taglib' marker file, check other PRs

timja (approved):

basil (approved):
Thanks for the PR!

timja:
@basil do you plan to release this soon?

Thanks!

basil:
Released in [2.1](https://github.com/jenkinsci/throttle-concurrent-builds-plugin/releases/tag/throttle-concurrents-2.1).

PR #100, opened 2020-12-12
Improved example for declarative pipelines in readme

StillLearnin:
I hunted for hours trying to figure out how to restrict concurrent builds.  When I finally found the answer on https://stackoverflow.com/a/63595178/934912 I thought it should be part of the documentation.


oleg-nenashev (approved):
+1, thanks!

basil (approved):
Thanks for the PR!

PR #117, opened 2021-03-12
[JENKINS-37809] Throttling based on parameters does not work with Pipeline jobs

basil:
Adds test coverage for this functionality for both classic job types and Pipeline jobs. The test for classic job types passed without the changes to `src/main`, but the test for Pipeline jobs failed without the changes to `src/main` due to [JENKINS-37809](https://issues.jenkins.io/browse/JENKINS-37809). With the changes to `src/main`, the test for Pipeline jobs now passes.


basil:
Released in [2.2](https://github.com/jenkinsci/throttle-concurrent-builds-plugin/releases/tag/throttle-concurrents-2.2).

PR #129, opened 2021-06-08
Support per-node limits on pipeline jobs

gpaciga:
This attempts to add support for per-node throttling of pipeline jobs, as discussed in https://issues.jenkins.io/browse/JENKINS-45140

In this initial attempt, I am getting throttle properties from pipeline jobs, and then changed how currently-executing jobs are counted.

When a pipeline job is started, before it even knows whether it can start, it occupies a one-off/flyweight executor. This should not count towards the throttle limit, or else it will always be off by one. After a pipeline starts, it continues to occupy a flyweight executor *and* one of the node's executors. We have to be careful about double-counting here too. For now I'm just excluding flyweights from the counts, but this may have unintended consequences. One expected flaw is that it will not count pipelines that aren't in a node{} block, but likely should.

So far I have only done some manual testing on scripted pipelines with node blocks. Tests have not been fixed/updated to support the new behaviour. If we can figure out what the appropriate logic is, I will need some help figuring how to do the tests.

Opening the PR despite the above flaws in the hopes of continuing the technical discussion of how this could work.

### DIFF
--- a/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
+++ b/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
@@ -454,6 +454,10 @@ public class ThrottleQueueTaskDispatcher extends QueueTaskDispatcher {
 
     @CheckForNull
     private ThrottleJobProperty getThrottleJobProperty(Task task) {
+        if (task instanceof PlaceholderTask) {
+            Job<?,?> p = (Job<?,?>) task.getOwnerTask();
+            return p.getProperty(ThrottleJobProperty.class);
+        }
         if (task instanceof Job) {
             Job<?,?> p = (Job<?,?>) task;
             if (task instanceof MatrixConfiguration) {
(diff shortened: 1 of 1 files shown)


gpaciga:
Thinking through the logic a bit more, the cases seem to be:

# a pipeline occupying a flyweight executor before it starts running should not be counted
# a pipeline occupying a flyweight executor should not be counted if it is also using a regular executor on a node (in which case, only the regular executor counts)
# a pipeline occupying a flyweight executor should be counted if it is not using a regular executor on a node

I don't know how we might distinguish between 1 and 3. Case 2 is tricky because the flyweight might be (or always is?) on the controller while the regular executor might be on a different node.

There is an alternative approach too: What if instead of saying pipelines can't be throttled per node at all, implemented it like this (or similar) and say that pipeline throttling only works when the job is using a regular executor (i.e. in a node block)? This is still a step up in functionality, and IIRC there is a best practice recommendation that pipelines should avoid running much code on flyweights anyway. Casual users probably don't even know that they exist. In fact, `pipelinesOnNode()` already excludes flyweights, so this behaviour seems consistent with how the plugin behaves elsewhere.

mamh2021:
good

PR #152, opened 2021-11-22
Remove usages of Guava

basil:


basil:
Released in [2.6](https://github.com/jenkinsci/throttle-concurrent-builds-plugin/releases/tag/throttle-concurrents-2.6).

PR #174, opened 2022-07-29
Add documentation and tests for throttle step in options block

nickb-carallon:
It is possible to use the throttle step in the options block of a
stage to throttle single stages in a declarative pipeline. This
commit documents this behaviour and adds a test for it.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -179,6 +179,29 @@ pipeline {
 }
 ----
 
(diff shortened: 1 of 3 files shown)


basil (approved):
Thanks for the PR!

nickb-carallon:
Oops, looks like I missed the separator at the end of my code block. Sorry!

PR #269, opened 2024-10-03
bugfix:Restrict Concurrency based on Runs

dharanikesav:
We have jenkins WorkFlowJobs that does not run on any specific nodes. When we run those pipelines, "Prevent multiple jobs with identical parameters from running concurrently" option is not working properly. To handle this scenario, have updated the logic to check ongoing builds based on runs in addition to nodes -> executors. 

This change is critical for us.

### Testing done

- Have not written any test cases for this

### DIFF
--- a/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
+++ b/src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
@@ -346,14 +346,13 @@ public class ThrottleQueueTaskDispatcher extends QueueTaskDispatcher {
                 // Checks if the build is currently in progress (could be waiting on input or paused)
                 if (run.isInProgress()) {
                     List<ParameterValue> runParams = getParametersFromRunItem(run);
-                    if(runParams.size() > 0) {
+                    if (runParams.size() > 0) {
                         runParams = doFilterParams(paramsToCompare, runParams);
                     }
 
-                    if(itemParams.containsAll(runParams)) {
+                    if (itemParams.containsAll(runParams)) {
                         return true;
                     }
-
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


basil:
I am not really maintaining this plugin anymore, so feel free to take over:

https://www.jenkins.io/doc/developer/plugin-governance/adopt-a-plugin/

dharanikesav:
Hi @basil , 
I am happy to take over. Have created a pull request for permission updater repo -> https://github.com/jenkins-infra/repository-permissions-updater/pull/4121

Please verify and approve. So that i can takeover from here.

Thanks,
Dharani

src/main/java/hudson/plugins/throttleconcurrents/ThrottleQueueTaskDispatcher.java
```diff
@@ -320,6 +325,40 @@ private boolean isAnotherBuildWithSameParametersRunningOnAnyNode(Queue.Item item
...
+        List<String> paramsToCompare = tjp.getParamsToCompare();
+        List<ParameterValue> itemParams = getParametersFromQueueItem(item);
+
+        if (paramsToCompare.size() > 0) {
+            itemParams = doFilterParams(paramsToCompare, itemParams);
+        }
+
+        if (item.task instanceof Job) {
+            Job<?, ?> project = (Job<?, ?>) item.task;
+            for (Run<?, ?> run : project.getBuilds()) {
```
basil, line 345: :bug: This will load every build of a project from disk, which is a huge performance hit and not appropriate for a hot method like this. Use `FlowExecutionList.get` instead to get active Pipeline jobs in memory. See also `ExecutorStepExecution#getOwnerExecutable`.

basil (changes requested):
No description of automated or manual testing.

PR #272, opened 2024-10-09
fix:improve concurrency restriction

dharanikesav:
We have jenkins WorkFlowJobs that does not run on any specific nodes. When we run those pipelines, "Prevent multiple jobs with identical parameters from running concurrently" option is not working properly. To handle this scenario, have updated the logic to check ongoing builds based on runs in addition to nodes -> executors.

This change is critical for us.

Testing done
Manual testing done locally for following test cases:
- Check if concurrency works when no parameters defined
- Check if concurrency works when parameters defined but throttle build not configured to be restricted based on parameters
- Check if concurrency work as expected when restricted based on parameters

cc: @basil


basil:
Is this problem not solved by #232?

dharanikesav:
Hi @basil , 

Sorry, i have not noticed the other PR. Yes, the issue is resolved now.

Thanks,
Dharani

PR #302, opened 2025-03-06
Migrate tests to JUnit5

strangelookingnerd:
This PR aims to migrate all tests to JUnit5. Changes include:

* Migrate annotations and imports
* Migrate assertions
* Remove public visibility for test classes and methods
* Minor clean up

There is one exception where a migration was not possible as the tests rely on `Rule` implementations that have not (yet) been migrated to JUnit5:

* `ThrottleJobPropertyPipelineRestartTest` using `JenkinsSessionRule` (in combination with `SemaphoreStep`)

I also took the freedom of re-enabling `ThrottleConcurrentTest` and refactor it to no longer rely on `jgiven`. 
IMHO there is no need for this test in particualar to use it and its better to streamline it with other tests in this plugin.

I am well aware that this is a quite large changeset however I hope that there is still interest in this PR and it will be reviewed.
If there are any questions, please do not hesitate to ping me.

- [x] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side) and not your master branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub 
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -125,23 +125,6 @@ THE SOFTWARE.
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-support</artifactId>
     </dependency>
-    <dependency>
-      <groupId>com.tngtech.jgiven</groupId>
-      <artifactId>jgiven-junit</artifactId>
-      <version>2.0.2</version>
-      <scope>test</scope>
-      <exclusions>
-        <!-- Provided by Jenkins core -->
-        <exclusion>
-          <groupId>org.fusesource.jansi</groupId>
-          <artifactId>jansi</artifactId>
-        </exclusion>
-        <exclusion>
-          <groupId>org.slf4j</groupId>
-          <artifactId>slf4j-api</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
 
     <!-- Dependencies for test -->
     <dependency>
--- a/src/test/java/hudson/plugins/throttleconcurrents/TestUtil.java
+++ b/src/test/java/hudson/plugins/throttleconcurrents/TestUtil.java
@@ -1,8 +1,8 @@
 package hudson.plugins.throttleconcurrents;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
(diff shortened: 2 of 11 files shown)


basil (changes requested):
Incomplete:

> There is one exception where a migration was not possible as the tests rely on `Rule` implementations that have not (yet) been migrated to JUnit5: `ThrottleJobPropertyPipelineRestartTest` using `JenkinsSessionRule` (in combination with `SemaphoreStep`)

strangelookingnerd:
> Incomplete:

While it is possible to use `JenkinsRule#restart` for most cases it does not yet work with `SemaphoreStep`. It seems like a Semaphore is not properly restored when restarting Jenkins and thus a restarted build is waiting forever. I was looking into solving that but found some roadblocks on the way. In regards to https://github.com/jenkinsci/oidc-provider-plugin/pull/68#discussion_r1987661165 I’m not sure if those rare cases (as I rarely encountered them so far when migrating other plugins to JUnit5) make it worthwhile to investigate further. WDYT?

strangelookingnerd:
From what I understood https://github.com/jenkinsci/workflow-support-plugin/blob/a600813d04aac835e7e2bf62a8aaa775c4cd4437/src/test/java/org/jenkinsci/plugins/workflow/test/steps/SemaphoreStep.java#L60 uses the `JENKINS_HOME` as a `File` object for the key when restoring / checking the `Semaphore`. `JenkinsRule#restart` makes a copy of the current home folder and thereby has a different path when being restarted , breaking the logic described above.

basil (changes requested):
Incomplete:

> There is one exception where a migration was not possible as the tests rely on `Rule` implementations that have not (yet) been migrated to JUnit5: `ThrottleJobPropertyPipelineRestartTest` using `JenkinsSessionRule` (in combination with `SemaphoreStep`)

strangelookingnerd:
@basil Now all tests are migrated suing the latest JTH version

